Issue #80, opened 2016-01-23
Make our templates less verbose


Changaco:
Even after #52 our templates will still be overly verbose. I want something better, probably more than what Jinja can provide.

rohitpaulk:
Copying over from #82:

> I personally wouldn't prefer the line statement syntax in HTML files - the file structure ends up looking inconsistent. However, I'd support a complete change to something like [HAML (ruby)](http://haml.info/tutorial.html) (https://github.com/Rootbuzz/shpaml, maybe?)

Changaco:
shpaml is interesting, thanks. HAML looks wrong to me, too ruby-like. I think the ideal for me would be [Shakespearean Templates](http://www.yesodweb.com/book/shakespearean-templates).

techtonik:
The future is after web-components anyway, so rather than focusing on `HTML syntax libraries`, maybe it is worth to investigate `component syntax libraries` and static API checkers for binding them with backends (such as with Flux architecture).

Changaco:
This ticket is not about switching to new technologies. Liberapay happily lives "in the past": we don't use a JavaScript framework, we don't use a big backend framework, we do things ourselves using proven technologies, and we'll keep doing it that way in the foreseeable future.

PR #82, opened 2016-01-23
Jinja line statements

### LINKED ISSUE #52: Switch Jinja templates to line statements
Because they're easier to type (especially on an azerty keyboard): `% if ...` instead of `{% if ... %}`. [Documentation](http://jinja.pocoo.org/docs/dev/templates/#line-statements).

Any objections?

Changaco:
Closes #52. Please take a look and tell me if this change looks okay to you.

### DIFF
--- a/410.spt
+++ b/410.spt
@@ -2,15 +2,15 @@ from aspen.http import status_strings
 [---]
 title = _('Closed') if 'username' in request.path else '410'
 [---] text/html via jinja2
-{% extends "templates/base.html" %}
-{% block content %}
-{% if 'username' in request.path %}
+% extends "templates/base.html"
+% block content
+% if 'username' in request.path
     <p>{{ _("The account owner has closed this account.") }}</p>
-    {% if user.ANON %}
+    % if user.ANON
     <h2>{{ _("Are you the account owner?") }}</h2>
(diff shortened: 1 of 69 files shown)


Changaco:
@rohitpaulk May I have your opinion? If this PR was for Gratipay would you approve it?

techtonik:
What about IDEs support for this Jinja syntax?

Changaco:
@techtonik I don't know, I don't use IDEs, but if an editor wants to support highlighting templates markup in HTML files then it should make the syntax configurable.

rohitpaulk:
I personally wouldn't prefer the line statement syntax in HTML files - the file structure ends up looking inconsistent. However, I'd support a complete change to something like [HAML (ruby)](http://haml.info/tutorial.html) (https://github.com/Rootbuzz/shpaml, maybe?) 

 The PR in itself looks good though :)

Changaco:
I agree that it looks inconsistent, but let's go with it anyway while we consider better solutions in #80.

Changaco:
!m @rohitpaulk @techtonik

Changaco:
This PR also activated Jinja's line comments, but they were breaking templates and we weren't using them, so I deactivated them in 1ee1e48fdad3099f4ac33c9dd7f3bbb26de62333.

PR #88, opened 2016-01-27
Get Liberapay running on python3

Changaco:
Blocked by https://github.com/gratipay/aspen.py/issues/493.


Changaco:
http://pythonclock.org/ I'm pretty sure we'll land this PR before that clocks runs out :-)

Changaco:
This is ready. We'll keep the code compatible with both python 2 and 3 as long as we haven't migrated the production server (and a little while after that in case we need to go back).

PR #89, opened 2016-01-27
Fix python3 compatibility

Changaco:
Fixes #87. Aspen uses `simplejson`, so bring it back and update it. I've removed the caches on Travis, that should prevent it from not detecting problems.


Changaco:
Travis is green, merging.

Issue #72, opened 2016-01-22
Automatic schema migration of local DB


Changaco:
In `schema.sql`:

- [x] add a `CREATE TABLE meta` with 2 columns `key(text), value(json)`
- [x] `INSERT INTO meta (key, value) VALUES ('schema_version', 0)`

In `release.sh`:

- [x] compute the new schema version number (add 1 to the previous one taken from `sql/migrations/`)
- [x] update the version number in the `INSERT` statement in `schema.sql`
- [x] add an `UPDATE` to `branch.sql` to modify the schema version in the DB
- [x] after running `branch.sql` against production, move it to `sql/migrations/<version>.sql` instead of deleting it

In `wireup.py`:

- [x] compare the version number in the DB to the one in `sql/migrations/` and run the scripts to catch up if necessary

Changaco:
Now that the production DB is up I'm switching to using `branch.sql` and I want to implement this.

PR #131, opened 2016-02-07
get rid of suspicious suspicious and blacklist fix #60

### LINKED ISSUE #60: Get rid of is_suspicious
While writing the Terms of Service (https://github.com/liberapay/liberapay.org/issues/4) it occurred to me that blacklisting users without notifying them and with no time limit is very bad. #58 and #59 are how we should handle fraud.

MartinDelille:


Changaco:
@MartinDelille You should take a look at the changes I've made here. We can't modify `schema.sql` directly now that the production DB is up, we need to put changes into a `branch.sql` file (the README has a section about this).

Changaco:
I've added a couple more commits to clean up a few things.

Changaco:
Travis is green, but I'd like to implement #72 before merging and deploying this.

Changaco:
#72 is done, I'm merging this. !m @MartinDelille

Issue #143, opened 2016-02-09
Detect broken links in tests


Changaco:
I broke the community join/subscribe buttons in #138 (fixed in 5ad5a607732d3d8a306268f337c7916dc1a61c93), a link checker would have spotted that.

Changaco:
I added a broken link in #1286 (fixed in 46b7a28cc8951c9e99454b8ad3555ce212d4ad92), that's another bug that would have been detected by a link checker.

PR #144, opened 2016-02-10
Change Jinja's `undefined` class to a stricter one

Changaco:
This would have prevented the bug mentioned in #143 yesterday, and has already detected another one that this PR fixes.


Changaco:
I've pushed an additional commit to log to Sentry when in production instead of failing completely to render the page. A partially broken page is better than a 500 error page in production.

Issue #128, opened 2016-02-07
Show payment processing fees in About pages


Changaco:
We don't actually show them anywhere on the site!

Changaco:
This issue was reported by a user in the first ever email to the support address. :-)

MartinDelille:
This should be displayed in the FAQ right? 

Are [these fees](https://www.mangopay.com/fr/nos-tarifs/) correct?

Changaco:
Yes, but we want to use the `FEE_*` constants from `liberapay/billing/exchanges.py`. We should move them to `liberapay/constants.py`, that way they'll be automatically available in simplates as `constants.FEE_*` without having to import them.

Changaco:
In addition to putting the info in the FAQ, we should mention the fees in the Introduction page and link to the new FAQ entry for details.

Changaco:
We should also tell/remind users about fees when they're about to pay them, so on the `/wallet/pay{in,out}` pages.

PR #159, opened 2016-02-14
display fee in the faq (ref #128)

MartinDelille:
This is just the first part of the fix for #128, it shows the fix and var fees in the faq. I don't really get where you want to add it in the introduction page.


Changaco:
I've completed this.

PR #167, opened 2016-02-17
Fix get_takes_last_week

Changaco:
There's a stupid SQL syntax mistake in it.


Changaco:
Before:

spectacle j10857

Now:

spectacle t10857

Issue #54, opened 2015-12-29
Implement setting a total giving amount


Changaco:
Once a user sets a total giving amount, the values of their individual donations act like shares of the whole. For example with a total giving of 10, one donation of 0.50, and another of 2.00, the actual amounts would be 2.00 (0.5/2.5) and 8.00 (2/2.5) respectively.

trebmuh:
I think this is a really great idea which makes a lot of sense as beautifully stated @Phyks in https://github.com/liberapay/liberapay.com/issues/923.

Phyks:
I missed this issue when looking for a similar one. Definitely a huge :+1: for this with a ratio per donation! :)

Changaco:
The problem with setting a total giving amount is that it's detrimental to income stability (which is a core goal of Liberapay). As new creators join the platform and start receiving donations the income of creators who had already joined earlier can decrease.

boehs:
So I propose "pools", I can make my tech pool at 5$ a month, my charity pool at 10$, and then do the ratio idea above for each pool

Issue #45, opened 2015-11-24
Implement payin by bank wire


Changaco:
It's not as convenient as credit cards but [the fee is only 0.5% instead of 1.8% + €0.18](https://www.mangopay.com/pricing/pricing-worldwide/).

juliencarnot:
Sorry if this is a stupid question, I haven't researched the issue, but have you considered offering to wire money directly from my bank account to liberapay's account by publishing an IBAN and asking users to reference a key in the description field of the transfer, to reconcile the transfer with a user account and credit its wallet? Might be a good option for a 0% fee (at least for a lot of French/European contributors), even if that means going through the transfer interface of their own bank account and might induce some delay before crediting the wallet. Thanks for your time and great project anyways!

Changaco:
We can't legally hold our users' money in our own bank account, that's why go through Mangopay. A zero fee option would be nice, but I think 0.5% is low enough, don't you? :-)

juliencarnot:
Ok, got it! Nothing beats the "Power of Free", but 0.5% will be nice ! ;-)

Changaco:
This will be done soon, see #316 for details.

Changaco:
Deployed! Anyone wants to test it? @juliencarnot?

![liberapay-2016-05-16-payin-live](https://cloud.githubusercontent.com/assets/1581590/15284485/be62cbb4-1b51-11e6-8bb2-2c22517aa9e0.png)

juliencarnot:
Sure! I checked this morning and did not find it but I can see it now! Will let you know how it went for me!

juliencarnot:
Went through the process without any problem, but a few remarks:
* first screen: "bank wire to a Euro account" should be "bank wire from a Euro account"
* second screen: 
 * for each amount option, a different fee is listed, and none equates the 0.585% fee "advertised" on top of the dropdown list, and the fee is higher for €13.09 than for €9.81. A bit confusing. Might be a typo?
   * €2.03 (1.48% fee included)
    * €3.28 (0.91% fee included)
    * €6.55 (0.76% fee included)
    * €9.81 (0.61% fee included)
    * €13.09 (0.69% fee included)
  * What if I want to put more than 13.09 € in my wallet?     
My bank will usually take a working day to complete the wire transfer, so I'll let you know tomorrow if my wallet is still "empty".

[...]

PR #168, opened 2016-02-17
Help users pay lower card fees

Changaco:
We've recently started saying on the homepage that Liberapay is "economical", but we've been doing a bad job at helping users pay less in credit card fees. This PR improves the situation a little, by raising the minimum charge from €10 to €15 (which lowers the fee to just below 3%), and by proposing to users who don't give much yet to add more money from the start so that they'll pay less than 2% in fees.

The next steps in lowering fees are #45 (flat 0.5% fee) and #54 (if we know the total amount the user wants to give instead of only the amount they're currently giving, then we can propose higher charge amounts that result in lower fees).

### DIFF
--- a/liberapay/billing/exchanges.py
+++ b/liberapay/billing/exchanges.py
@@ -15,8 +15,8 @@ from mangopaysdk.types.money import Money
 
 from liberapay.billing import mangoapi, PayInExecutionDetailsDirect, PayInPaymentDetailsCard, PayOutPaymentDetailsBankWire
 from liberapay.constants import (
-    FEE_CHARGE_FIX, FEE_CHARGE_VAR, FEE_CREDIT, FEE_CREDIT_OUTSIDE_SEPA,
-    MINIMUM_CHARGE, QUARANTINE, SEPA_ZONE,
+    CHARGE_MIN, CHARGE_TARGET, FEE_CHARGE_FIX, FEE_CHARGE_VAR,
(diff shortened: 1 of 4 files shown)


Changaco:
Are there objections to raising the minimum charge from 10 to 15 euros? Technically speaking that's a 50% increase, but it's still seems low enough to me.

techtonik:
Are there any prooflinks that raising charge to €15 lowers the `credit card fee` to 3%? What does `credit card fee` mean exactly, by the way?

Changaco:
@techtonik You've been hanging around Gratipay for a while, you should know what fees are by now.

Liberapay's payment processor is MangoPay, and they charge 1.8% + €0.18 per credit card payment (it's not all for them though, I'm guessing a significant portion goes to intermediaries further down the chain). Mathematically, since the fee has a fixed part, you reduce the overall fee rate by making bigger payments. Illustration: to get €10 in your Liberapay account you actually have to pay €10.37, and 0.37 is 3.57% of 10.37, but if you up the amount to €15 then you have to pay €15.46, and 0.46 is only 2.98% of 15.46, so relatively speaking you've paid less in fees.

Changaco:
FTR Zatalyz answered "no objection" in IRC.

techtonik:
Well, fees are constantly changing - that's why I asked about the prooflink. `1.8% + €0.18` seems fair. Worthy to implement as a calculator on JS enabled browsers.

Changaco:
This PR already computes the overall fee rate server-side and shows it to the user.

PR #169, opened 2016-02-18
More pledges

Changaco:
We haven't been doing a good job at telling users that they can pledge to people who haven't joined Liberapay yet. This PR allows us to see who other users are pledging to, and advertises pledging on the homepage.

### DIFF
--- a/templates/explore.html
+++ b/templates/explore.html
@@ -2,6 +2,10 @@
 % block subnav
     % from "templates/nav.html" import nav with context
     <ul class="nav nav-tabs">{{
-        nav([('/', _('Communities')), ('/teams/', _('Teams'))], base='/explore')
+        nav([
+            ('/', _('Communities')),
+            ('/teams/', _('Teams')),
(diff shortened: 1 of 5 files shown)


Changaco:
Screenshot of the new Explore Pledges page:

liberapay-2016-02-17-explore-pledges

MartinDelille:
> Liberapay allows you to pledge to donate to people...

I'm not a native english speaker but there is a repetition here, isn't it a problem?

Changaco:
It is a lot of "to"s in one sentence and should probably be reworded. However this message is just copy-pasted from another page, it's not being added by this PR, and if we change it we have to update the translations. Moreover, "pledge to donate" is not exactly a repetition: "pledge" means "promise", if we don't say what's being promised then the sentence is slightly ambiguous.

Changaco:
This is deployed. Now let's make some pledges, because right now https://liberapay.com/explore/pledges/ is empty!

PR #170, opened 2016-02-18
Some UI fixes

Changaco:
See the commit messages.


Changaco:
Added a commit to fix an overflow at the bottom of receipts (they still suck though). Before:

spectacle tj5106

After:

spectacle xm5106

PR #178, opened 2016-02-22
Fix CORS

Changaco:
Deploying #176 revealed this issue.


Changaco:
I had to purge the cache on CloufFlare for this to take effect.

Issue #76, opened 2016-01-23
Add Gitlab as elsewhere platform


Changaco:
We've considered moving to Gitlab (#46) but we don't even have it as an elsewhere platform.

techtonik:
`elsewhere` cries to be a separate project for a long time. =)

Changaco:
Indeed, I just never got around to it, reticketed as #79.

PR #188, opened 2016-02-25
Add two more elsewhere platforms (GitLab.com and LinuxFr.org)

Changaco:
Implements #76 and [this suggestion from August 2014](https://linuxfr.org/users/changaco/journaux/presentation-de-gratipay-pour-un-financement-durable-du-logiciel-libre#comment-1558034).


Changaco:
I created https://redirect-to-http.oy.lc/ to work around LinuxFr's requirement that the callback URL be HTTPS.

PR #189, opened 2016-02-26
Move app configuration to the database

### LINKED ISSUE #181: Move app configuration from environment to database
This wouldn't accomplish #150, but it would reduce the severity of the problem by bypassing it.

Other advantages:

- it would allow detecting legacy variables that haven't been removed, though we could also achieve that with environment variables by prefixing their names
- we should be able to drop `environment` from our dependencies, and maybe `honcho` too
- our DB backups would contain the app's configuration

Drawbacks: ?

Changaco:
Closes #181.


Changaco:
While I was at it, I did a little more cleanup.

PR #225, opened 2016-03-16
Fix asset caching

Changaco:
We still had problematic race conditions and corner cases in there, this PR should fix them.


Changaco:
Travis is green. I'm merging now because running payday will be less problematic with this fixed.

PR #226, opened 2016-03-16
Fix handling of unfunded donations to teams in payday

Changaco:
https://github.com/liberapay/salon/issues/24#issuecomment-197300394


Changaco:
Ready for review. We don't need this merged until next week, so there's time.

PR #233, opened 2016-03-23
Make sure we have a recent version of virtualenv

### LINKED ISSUE #232: make env errors out
```bash

"/usr/bin/python2.7" -c "import virtualenv" || pip install virtualenv
"/usr/bin/python2.7" -m virtualenv --no-download ./env/
Usage: virtualenv.py [OPTIONS] DEST_DIR

virtualenv.py: error: no such option: --no-download
make: *** [env] Error 2
```

Changaco:
This is an attempt to fix #232. @aandis can you try that?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ install_where := $(shell $(python) -c "import sys; print('' if hasattr(sys, 'rea
 # NOTE: Creating a virtualenv on Windows places binaries in the 'Scripts' directory.
 bin_dir := $(shell $(python) -c 'import sys; print("Scripts" if sys.platform == "win32" else "bin")')
 env_bin := env/$(bin_dir)
+env_py := $(env_bin)/python
 test_env_files := defaults.env,tests/test.env,tests/local.env
 pip := pip --disable-pip-version-check
 with_local_env := $(env_bin)/honcho run -e defaults.env,local.env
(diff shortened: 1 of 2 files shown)


Changaco:
Added some commits to fix Travis and a few other things.

aandis:
Hm. I don't have `ensurepip` either. 

```bash
"/usr/bin/python2.7" -m ensurepip --user
/usr/bin/python2.7: No module named ensurepip
make: *** [env] Error 1
```

Installing..
Probably better to add it as a dependency in README if we are gonna go this route.

Changaco:
[`ensurepip` is a built-in module](https://docs.python.org/2/library/ensurepip.html), it was added in python 2.7.9 / 3.4 (I hadn't realized that, I thought we could count on it always being available).

I've added a commit to show a warning instead of failing when `ensurepip` is missing.

aandis:
arghhh. Now I get..

```bash
"/usr/bin/python2.7" -m ensurepip --user || make --no-print-directory _warning
/usr/bin/python2.7: No module named ensurepip
-e 
Warning: you're using an old version of python, you really should upgrade!

"/usr/bin/python2.7" -m pip --disable-pip-version-check install --user "virtualenv>=15.0.0"

Usage:   
  /usr/bin/python2.7 -m pip <command> [options]

no such option: --disable-pip-version-check
make: *** [env] Error 2
```

I am just gonna go ahead and update all my dependencies to the latest version eh?

Changaco:
Yeah, there's only so much backward compatibility we can support.

BTW, what OS are you using?

aandis:
Ubuntu 14.04

PR #236, opened 2016-03-24
Encode b64decode_s input to utf-8. Fixes #235

### LINKED ISSUE #235: b64decode_s with unicode crashes.
Was working on https://github.com/gratipay/gratipay.com/pull/3965 when I ran into this.
According to http://stackoverflow.com/a/2230623/2643846

> b64decode quite explicitly can only take bytes (a string), not unicode.

Steps to reproduce - 

1. `b64encode_s('TheEnter?prise')` => `VGhlRW50ZXI_cHJpc2U~`
2. `b64decode_s('VGhlRW50ZXI_cHJpc2U~')` => `TypeError: character mapping must return integer, None or unicode` (deep down in the stack trace)

aandis:


liberapay/utils/__init__.py
```diff
@@ -109,6 +109,8 @@ def get_community(state, restrict=False):
 
 def b64decode_s(s, **kw):
     udecode = lambda a: a.decode('utf8')
+    if not isinstance(s, bytes):
+        s = s.encode('utf8')
```
Changaco: The appropriate codec here is `'ascii'`, the input isn't supposed to contain any non-ascii character.

aandis: Oh is it? What is actual reason for 11b52fd30273cc213a9d4a5e8a86ee55615b386a to fail then?

aandis: As in, what non-ascii characters does `VGhlRW50ZXI_cHJpc2U~` have?

Changaco: I only meant that the `s` string here isn't supposed to contain any non-ascii char, so the code would be clearer if we used `'ascii'` instead of `'utf8'`.

aandis: Cool. Fixing.

aandis:
@Changaco fix pushed. Check them out.

Changaco:
!m @aandis :-)

PR #245, opened 2016-04-05
Run tests in parallel

Changaco:
It'd be nice if our tests ran faster, this PR tries to achieve that through parallel processing. Unfortunately, our tests are not designed for this, each process acts as if it's the only one using the test database. We should try to change that, but this is low-priority DevX work, so I'm putting it aside for now, feel free to assign this PR to yourself if you want to work on it.

Here are some relevant links I found after a quick search:
- http://dba.stackexchange.com/questions/76494/temporary-schema-per-connection
- https://pypi.python.org/pypi/testing.postgresql/


whit537:
Suggest closing as stale to clear out the PR queue.

Issue #239, opened 2016-03-30
Payout status isn't updated properly


Changaco:
The `exchanges.status` column of withdrawals is not being updated to `succeeded` even though the payouts have been carried out successfully.

Issue #249, opened 2016-04-06
Payin failures aren't handled properly


Changaco:
Background: https://github.com/liberapay/salon/issues/30.

PR #250, opened 2016-04-06
Fix handling of payment failures

Changaco:
I think I've found the cause of #249: we assume the transaction has succeeded if an exception isn't raised, when in fact it looks like the mangopay library doesn't raise an exception for simple payin failures.


Changaco:
We make the same assumption for payouts, I've modified that code as well. This should fix #239.

PR #255, opened 2016-04-07
Fix update_cached_amounts

Changaco:
The current code doesn't update participants who weren't active during payday.


Changaco:
Somehow it's now even more broken than before… /o\

PR #256, opened 2016-04-07
Fix update_cached_amounts - part 2

Changaco:
Completes #255.


Changaco:
Better, but still slightly off. :-/

PR #257, opened 2016-04-07
Fix update_cached_amounts - part 3

Changaco:
Fixes #256.


Changaco:
Found a solution that should do it for now, though we may want to revisit how `giving` is computed in the future.

Changaco:
`receiving` is still off for the Liberapay team. :-/

I've been using this to fix it: `for p in website.db.all("select p.*::participants from participants p where kind = 'group'"): p.update_receiving()`

Changaco:
`receiving` is still broken: it's `0.00` for teams that don't distribute their income.

PR #275, opened 2016-04-23
Show the amount the user gives in EUR

astiob:
It is currently shown in USD. As far as I can tell, this is the only user-visible place where USD appears.

### DIFF
--- a/www/%username/giving/index.html.spt
+++ b/www/%username/giving/index.html.spt
@@ -53,7 +53,7 @@ weekly = total - participant.receiving
         % endif
     </p>
 % else
-    <p>{{ _("You give {0} per week.", Money(participant.giving, "USD")) }}</p>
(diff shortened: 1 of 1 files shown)


Changaco:
Good catch, and welcome to Liberapay. :-) Would you like to join the team?

Changaco:
Fix deployed.

astiob:
> Would you like to join the team?

Ha ha, thanks for the offer! But I’m not sure I’m ready to take on additional responsibilities at the moment.

Changaco:
@astiob As you wish. To clarify, it's not really a responsibility, it can be purely honorary: we currently have 4 members who are on the team but [don't actually take money from it](https://liberapay.com/Liberapay/income/), and most of them haven't worked on the project in a while.

PR #303, opened 2016-05-04
Don't send "identity required" emails to teams

Changaco:
A user has reported that he received such an email, that's a bug.


Changaco:
I've reviewed the other payday notifications, I don't think they have the same issue.

Issue #312, opened 2016-05-11
Code Analysis


sils1297:
Hey,

we're coala and are using your service. Would you be interested in getting static code analysis for your whole codebase? We can help you improving your code style in one go (coala can do some fixes automatically) and/or improving code style and quality incrementally (GitMate is a code review bot we're building that will only show new issues).

Changaco:
PRs to improve Bug Prevention are very welcome. :-) We currently only have pyflakes partially covering our codebase, it would be great if we could also analyze simplates (#122), and our JavaScript code.

sils1297:
CC @sims1253 @SanketDG

Changaco:
A useful analysis of JavaScript would be undefined variable detection (#100). We'd need to analyze the entire JS code at once for that, not each file individually.

sils1297:
Yeah I looked into that but our routines are too stupid for your setup without anything custom. For that we'd need to compile it together, then do the analysis I guess?

Changaco:
We have a python function that compiles our JS and CSS, we could call it before coala, and point the analyzer to `www/assets/*.js`.

sils1297:
Cool, what command is it? Is there a make target for that?

Changaco:
We don't have a command to compile the assets, and the python code automatically cleans them up on exit so we'll need to tweak that.

sils1297:
Ok, which function is it and which function cleans it up? We can do a make target for coala that compiles it and then runs the analysis.

Changaco:
It's all in `liberapay/utils/http_caching.py`. The `compile_assets` function automatically removes files that didn't exist before it created them. I think we should stop doing that and add a `atexit.register(lambda: clean_assets(www_root))` line back to `liberapay/testing/__init__.py` where it used to be (eef2932332f0ae5cd08815c6b19ad3872c762804).

PR #314, opened 2016-05-11
Add coala configuration

sils1297:
Run Python analysis only with `coala py-analysis` like a make target.
Install coala with the official analysis routines via:

    pip3 install coala-bears

The output of coala-ci (noninteractive human readable result dump
intended for continuoue integration) is available at
http://pastebin.com/VGHv5E7K

See https://github.com/liberapay/liberapay.com/issues/312 .

I added analysis sections only as you didn't seem to be too much interested in codestyle analysis? Semantic issue detection is indeed a problem due to your templating. I wonder if we can solve that language independently at a coala level...


sils1297:
So this is duplicate code detection and complexity measurement only.

Changaco:
I'm not opposed to code style analysis, as long as it's properly configured to not raise lots of warnings we don't care about. Nobody's ever taken the time to set that up.

sils1297:
We can automatically reformat your code so you don't get tons fo warnings.

Changaco:
I'd like to see a diff of the automatic changes that would be applied to our current code before agreeing to turn that feature on.

sils1297:
Of course, coala never touches your code unless you ask it to. I can initially make a PR fixing up your Python code to PEP8 e.g. and from then one we can have GitMate comment on style violations on GitHub PRs if you want?

Changaco:
What we need to do here is integrate coala into our Makefile so that developers and CI will use it seamlessly. I don't think we want PR comments, errors should simply show up as CI build failures.

sils1297:
Well, right now it yields a lot of failures and thus your CI would fail. You can configure GitMate to only make a commit status for every commit and not comment on the PR, advantage here is that GitMate is smart enough to only see new issues which is hard to do in your CI because you have to run coala over several commits for that.

sils1297:
Added codestyle commits, assuming maximum line length of 120 for python

Changaco:
- We can drop the spacing check, we don't want to modify the files it modified.
- pep8 is making many changes we don't really want, and when it's detecting real anomalies it's not always fixing them properly. Can we tweak the configuration?

sils1297:
Yes, we can tweak the config. coala can generate a documentation for every analysis when passed the `-B` argument: http://pastebin.com/4EmqhHev

Changaco:
One of coala's dependencies (`language-check-0.7.6`) is currently failing to install on my machine. Also, coala is a bit heavy, and being python3-only complicates the integration into our Makefile. Maybe we should wait a while, try again later this year?

sils1297:
Sure! coala is rather heavy ATM because it installs the full set of analyses which covers almost 40 programming languages. We have a GSoC student working on seperating it into packages so you can only take what you need after the summer.

Changaco:
Awesome! Thanks for all your work @sils1297. I look forward to Liberapay using coala this autumn. :-)

Changaco:
In the meantime I've made a PEP8 PR: #333.

PR #327, opened 2016-05-17
Allow team members to take more when there is unused income

### LINKED ISSUE #287: Tweaking the take throttling algorithm
We're trying to use liberapay for our catalyst community in Lille, that's why we got 100€ (https://liberapay.com/collectifcatalyst/) by week for our local community. We need people to take this money faster than starting with 1€ (and 2€ / 4 / 8 next week) as people are giving a lot of time now. Is that possible ? Could we get multiple way to distribute money in a team depending on how the project go (for example for a faster project, starting with 1€ is difficult).

Changaco:
This PR will close #287.

### DIFF
--- a/liberapay/models/_mixin_team.py
+++ b/liberapay/models/_mixin_team.py
@@ -75,7 +75,6 @@ class MixinTeam(object):
           ORDER BY member, mtime DESC
 
         """, (self.id,)) if t.amount}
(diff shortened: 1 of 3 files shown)


Changaco:
This is far from a perfect solution, but let's go with it anyway.

PR #345, opened 2016-06-01
Invert donors pitch

MartinDelille:
After showing the plateform to people, it seems the think that Liberapay is a software thing. I just invert the two paragraph so people see directly what it is, far more generic that free software development. I'm ok to discuss this change so feel free to comment!

### DIFF
--- a/templates/pitch-donors.html
+++ b/templates/pitch-donors.html
@@ -1,12 +1,12 @@
 <p>{{ _(
-    "People who contribute to the commons need you to support their work. "
-    "Building free software, spreading free knowledge, these things take "
-    "time and cost money, not only to do the initial work, but also to "
-    "maintain it over time."
-) }}</p>
-<p>{{ _(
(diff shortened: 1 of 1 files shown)


Changaco:
Simply reversing the two paragraphs is awkward, especially since the one you want to put first ends with a colon.

MartinDelille:
Yes I did this quickly yesterday. I'm trying to address the critic that was made from people who tried to access the plateform, but it leads to a more fundamental question: who is Liberapay designed for? only creator? does it fit for activist? someone making helpdesk support? Do we need to target a particular audience (creators) or make it as generic as we can?

Changaco:
Liberapay is for anyone who wants to use it, but our primary goal is to fund free software, because that's what Liberapay is and we want it to fund itself. Our secondary user base is all other creators of commons, whether they make educational content, entertainment, open source hardware, or something else.

PR #349, opened 2016-06-06
Improve fee computation

Changaco:
This PR modifies how we compute fees to try to get closer to the actual amounts.

Modifying the fee formula is dangerous, so I ran a simulation to validate the soundness of the change, and it suggests that this would have reduced this month's surplus from €0.39 to €0.26, and last month's from €0.04 to €0.01. (I announce the surplus every month in https://github.com/liberapay/liberapay.org/issues/10.)


Changaco:
Here's what I used for the simulation: https://gist.github.com/Changaco/3dc4d5f0265f921847a8bc6e66c0e92d

PR #370, opened 2016-07-07
Fill `Tag` field of MangoPay Users with Liberapay profile URL

Changaco:
MangoPay requested this in order to "verify the conformity" of our users.

### DIFF
--- a/www/%username/identity.spt
+++ b/www/%username/identity.spt
@@ -34,6 +34,9 @@ if request.method == 'POST':
         p = 'LegalRepresentative'
         keys = KEYS_LEGAL
 
+    if not account.Tag:
+        account.Tag = website.canonical_url + '/~%i/' % participant.id
+
     account.Email = participant.email
     for k in keys:
         v = body.get(k)


Changaco:
Here's the script I'm planning to use to backfill the `Tag` for existing users:

```python
from __future__ import division, print_function, unicode_literals

from liberapay.billing import mangoapi
from liberapay.main import website

participants = website.db.all("""
    SELECT p.*::participants
      FROM participants p
     WHERE mangopay_user_id IS NOT NULL
""")

for p in participants:
    tag = website.canonical_url + '/~%i/' % p.id
    account = p.get_mangopay_account()
    if account.Tag != tag:
        print("setting Tag for user #%s (liberapay id %s)" % (account.Id, p.id))
        account.Tag = tag
    else:
        print("skipping user #%s (liberapay id %s) because Tag is already set" % (account.Id, p.id))
        continue
    for k, v in list(account.__dict__.items()):
        if k.endswith('Address') and not getattr(v, 'Country', None):
            setattr(account, k, None)
    account = mangoapi.users.Update(account)
    assert account.Tag == tag
```

Changaco:
The script posted above has been run successfully.

PR #373, opened 2016-07-08
Fix cash bundles generated by refunds

Changaco:
This PR adds a DB check that detects several inconsistencies not covered by the existing tests, and will fix our code accordingly.

### DIFF
--- a/liberapay/billing/exchanges.py
+++ b/liberapay/billing/exchanges.py
@@ -418,6 +418,7 @@ def propagate_exchange(cursor, participant, exchange, route, error, amount):
 
     if amount != 0:
         participant.update_giving_and_tippees(cursor)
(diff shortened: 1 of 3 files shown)


Changaco:
P.S. to develop the DB check in this PR I ran a simple `db-checks.py` script on the production server, it looked like this:

```python
from __future__ import division, print_function, unicode_literals

from liberapay.main import website

# Run the existing checks first
website.db.self_check()

# Then experiment with new checks...
l = website.db.all("""
   [...]
""")
assert len(l) == 0, "bundles and/or e2e_transfers are out of whack:\n" + '\n'.join(str(r) for r in l)
```

Changaco:
aa08b55b3a165eff8350a7815c62614b6a0fd22e completes this PR.

Issue #377, opened 2016-07-18
Unable to send IBAN information by mail


MartinDelille:
I decided to add money via IBAN transfert. When I got to this address: https://liberapay.com/martindelille/wallet/payin/bankwire/?exchange_id=XXX&payin_id=XXXXXX, there is a button proposing me to send me the information by email.

When clicking on it I have the following error message (not translated into french):

    Looks like you've found a bug! Sorry for the inconvenience, we'll get it fixed ASAP!

Changaco:
Seems to be a transient network connectivity issue, just try again.

MartinDelille:
After several tried it worked! Maybe the error message could be updated and translated?

MartinDelille:
What do we need to make this string translatable with weblate? https://github.com/liberapay/liberapay.com/blob/55d63f8b8c009f688e4189308a42fc619dd9b7b8/error.spt#L23

Changaco:
No, it can't be translated, it's the last error message we show when everything fails miserably, i18n is not available at that point.

MartinDelille:
Ok

Changaco:
Oh wait, I was thinking of that one: https://github.com/liberapay/liberapay.com/blob/master/liberapay/main.py#L82

The one you saw is already supposed to be translated.

Changaco:
We need to add `error.spt` to the list of files we extract messages from: https://github.com/liberapay/liberapay.com/blob/master/Makefile#L75

PR #378, opened 2016-07-18
Add error.spt to translatable files

MartinDelille:
After discussion here: #377


Changaco:
```
Usage: pybabel extract [options] dir1 <dir2> ...

pybabel: error: 'error.spt' is not a directory
```

/o\

MartinDelille:
I will revert it directly on master and try to propose a better way to do it.

Issue #183, opened 2016-02-23
Making the Liberapay site visually attractive


Calinou:
Currently, the Liberapay site does not have much identity. It uses the default Bootstrap 3 theming, which is quite bland. Upgrading to Bootstrap 4 (which will be released possibly this year) will make it look a bit more "current", but it still won't give the project much identity.
## What could be improved
- Images/gradients 
  - Primarily as "jumbotron" background, to give the site a more visual feel. Of course, the image shouldn't be too heavy in file size.
- Fonts
  - Font families, sizes, rendering
- Colors
  - To give the site more identity: needs a color palette first. This brings us the question, "what would be the color(s) of Liberapay?"
- Animations and transitions
  -  Some subtle animations make for a better user experience, telling the user where to go more clearly

I could work on this if desired. I have some HTML/CSS experience, having worked on sites like [minetest.net](http://www.minetest.net/) or [godotengine.org](http://godotengine.org/).

For future reference, here's a screenshot of the current website:

![Liberapay.com screenshot](https://lut.im/nvoUrqKSkv/BlWOYK5FFknL78NY.png)

Changaco:
Hey @Calinou, welcome. :-)

I agree with all your improvement ideas:

- Yes, Liberapay would be better with some kind of background illustration to put on the homepage and on our twitter/facebook/etc accounts. I like [Snowdrift](https://snowdrift.coop/)'s animated drawing, but Liberapay is more generic and I'm not sure how to explain it with a drawing. The comic at the top of [this blog post](https://medium.com/@nayafia/what-success-really-looks-like-in-open-source-2dd1facaf91c) is also a nice illustration of the primary problem Liberapay is trying to solve. Another possibility is images of people happily working in shared spaces or outside in a sunny day, evoking team work and freedom, though that's all a bit stereotypical.
- Yes, a less-widely-used font could be useful. I'll probably only agree to switch to a free one though.
- Yes, we need a colour palette, unfortunately I still have no idea. Feel free to make propositions.
- I'm fine with adding some animations/transitions, I hadn't thought of that.

Although I don't think insufficient visual identity is Liberapay's biggest problem right now, help on all this is indeed desired, so feel free to contribute. ;-)

Changaco:
Screenshot of Snowdrift's homepage illustration, FTR:

[...]

PR #387, opened 2016-07-25
Integrate Bootstrap

Changaco:
This PR is to unblock #183, it allows customizing Bootstrap by modifying the new `style/variables.scss` file. Ping @MartinDelille and @Calinou.


Changaco:
TODO:

- [ ] fix the tests
- [ ] update the CSS section of the README

Changaco:
Coincidentally, [Bootstrap 3.3.7 has just been released](http://blog.getbootstrap.com/2016/07/25/bootstrap-3-3-7-released/), so we should upgrade before merging this.

MartinDelille:
This is awesome and will help us make it look nice :-)

Just a question: why not using git submodule to link bootstrap code?

Changaco:
> Just a question: why not using git submodule to link bootstrap code?

It didn't come to mind because I never use them and they have a bad reputation, but I guess we could try.

MartinDelille:
It is not a big deal when you use it for external projects. A lot of bug have been fixed also. After that I think bower.io is also a simple solution for referencing external code (simpler to use than git submodule). Any preference?

Changaco:
No Bower. Preferred solutions:

1. git submodule if it's not a pain in the ass
2. a tarball stored with git-lfs and unpacked at the appropriate time
3. just leave things as they are now

MartinDelille:
Let's go with 3, but using submodule afterward will make handling upgrade easier.

Changaco:
We can make it easy to upgrade no matter which option we choose.

MartinDelille:
Maybe, but using submodule reduce the codebase and the diff. Although it is always possible to patch a vendor code by forking it and referencing the fork in the submodule.

Changaco:
Option 2 (using a tarball and git-lfs) has the same advantages, only option 3 (storing the files directly in our git) generates diff noise.

MartinDelille:
You're right. I never used git-lfs yet.

Changaco:
I explored the possibilities of using git-lfs for this, but if we did that then we'd have to add git-lfs as a required dependency (right now it's only used for the PDFs of MangoPay's terms, so it's optional).

Changaco:
Deploying this broke everything because `libsass` failed to compile with the obsolete version of `gcc` present on the server.

Changaco:
Upstream has a pull request to bring back compatibility with older GCC versions, https://github.com/sass/libsass/pull/1623, but we can't wait for that.

Changaco:
The libsass python package actually provides binaries. `pip` wasn't using them because it was also out-of-date. After upgrading pip, installing an up-to-date precompiled version of libsass succeeded, and re-deploying version 126 of liberapay worked.

MartinDelille:
Regarding the submodule option, usually, I create a vendor directory in the root directory. Do you prefer me to keep the current architecture and I will have to find how to make a submodule that point to a repository subdirectory (https://github.com/twbs/bootstrap-sass/tree/master/assets/stylesheets).

What do you prefer? If you think it is a waste of time, tell me.

Changaco:
There's no point in changing the integration now, we went with option 3 (directly including the Bootstrap source in our git), let's stick with it until we have a good reason to switch to something else.

Issue #46, opened 2015-12-04
Migrate to gitlab.com


Changaco:
[GitLab](https://about.gitlab.com/) has been continuously [improving](https://about.gitlab.com/2015/11/22/gitlab-8-2-released/), and it should now be possible to migrate Liberapay with only a small effort. The advantages are: it's open source, the gitlab.com instance provides free private repositories, GitLab is [a remote company](https://about.gitlab.com/2015/04/08/the-remote-manifesto/) (we're a remote organization). The downsides are: need to migrate from Travis CI to GitLab CI, smaller user community than GitHub.

So, should we migrate?

Changaco:
The missing [watch a repo](http://feedback.gitlab.com/forums/176466-general/suggestions/5427278-watch-a-repo) feature is a blocker. Closing this issue for now, we'll reconsider later.

sim6:
:(

Changaco:
GitLab issue for the "watch a repo" feature: https://gitlab.com/gitlab-org/gitlab-ce/issues/9013

Changaco:
It looks like watching a repo will be implemented soon, it's time to reconsider this.

Changaco:
Watching a repo has been implemented by https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/3986.

Changaco:
Someone told me in IRC that GitLab CI is great but that you have to set up your own worker, which is a bit of additional work compared to simply using Travis.

EliotBerriot:
We have a full Gitlab / Gitlab CI installation at work, It's a bit painful to setup the first time, but not too much. I'm okay to help with that if you want (I'm managing my workplace's Gitlab since almost 2 years).

Changaco:
I don't think we want our own Gitlab instance unless using gitlab.com turns out to be problematic.

EliotBerriot:
If you use the hosted version, I think you can use also their CI workers, (meaning you just need to write relevant configuration in `.gitlab-ci.yml`), it should be easy enough to set up. Maybe it's also possible to have an hybrid setup: Hosted Gitlab, self-hosted CI worker ?

Changaco:
> Maybe it's also possible to have an hybrid setup: Hosted Gitlab, self-hosted CI worker ?

It's almost certainly possible, but we should try the shared CI first.

P.S. in English there are no spaces before punctuation marks. ;-)

EliotBerriot:
[...]

PR #388, opened 2016-07-27
Add gitter badge to README.md

MartinDelille:
After liberapay/salon#58, this is a proposal to add gitter as a communicating plateform. I don't have right to create the room, so we need to create it first.

- [ ] Create Gitter room

I'll be happy to answer question to people in the chat room!

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,11 @@ It's a fork of Gittip/Gratipay, see [this post](https://github.com/liberapay/sal
 
 ## Contact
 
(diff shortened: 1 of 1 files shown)


Changaco:
-1 from me, because it's not good to have lots of communication channels, and Gitter isn't open source and is tied to GitHub (which we want to move away from eventually, see #46).

MartinDelille:
Ok no problem. What do you think of https://www.mattermost.org then?

PR #402, opened 2016-08-30
correction

finkiki:
corrige une faute


Changaco:
Thanks, but please make translation fixes on Weblate (a simple search finds two strings that need to be fixed: <https://hosted.weblate.org/translate/liberapay/core/fr/?q=rester+informer&search=substring&source=on&target=on&type=all&ignored=False>) instead of opening pull requests here.

PR #422, opened 2016-10-14
Add warning for bank SWIFT fee

revi:
When I was receiving funds from liberapay, I received a call from my bank's personnel saying "you have to pay a fee[1] but we'll waive the fee for this time". I think this is worth noting that your bank may have a fee on (even) receiving funds.

And well, I'm not sure about indentation or any other place that need same notice so review is really welcome.

[1]: fee was greater than the total receiving amount, fyi.


Changaco:
Looks good to me. Thanks @revi! :-)

techtonik:
> [1]: fee was greater than the total receiving amount, fyi.

Yea, tracking bank fees would be a killer feature.

PR #431, opened 2016-11-04
Add support for montly and yearly donation amounts

### LINKED ISSUE #85: Support monthly and yearly amounts in donation form
I'm not sure yet how this should be implemented.

Changaco:
This PR closes #85.

In reality donations are always processed weekly (payday isn't modified), but now users can input a monthly or yearly donation amount and the server automatically computes the weekly equivalent (see `PERIOD_CONVERSION_RATES`) which is then used during payday.

Each donation has its own period, so you can have any combination of weekly, monthly, and yearly donations. This may be unnecessary complexity, perhaps one period per user would have been simpler, I'm not sure.

### DIFF
--- a/js/10-base.js
+++ b/js/10-base.js
@@ -52,12 +52,18 @@ Liberapay.init = function() {
         }
     });
 
-    var amount_re = /\?(.*&)*amount=(.+?)(&|$)/;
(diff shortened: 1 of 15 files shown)


Changaco:
Ready for review (anyone available?). Here's what a Donate page looks like on this branch:

spectacle m12917

PR #449, opened 2016-11-30
Fix #448: added Docker configuration for easier development

### LINKED ISSUE #448: Development environment using Docker and docker-compose
I fell like I want to hack on this project, however, considering the requried setup (postgresql, virtualenv, python dependencies, etc.), I think I'm gonna experiment a dev setup using Docker.

If you are not familiar with it, Docker and especially [docker-compose](https://docs.docker.com/compose/) are tools you can use to containerize an application, effectively isolating it from the host machine. Docker-compose takes care of running all services required for the project (i.e. a Postgres database, a cache server, a SASS compiler, etc.), with a single command.

It's especially useful during development (although you can deploy apps using Docker, too), because you can spin up a development environment in minutes, with a single command, and without polluting your host machine with databases, web servers and every service any project requires.

Also, containers hide the host machine details and make installation a lot simpler: some packages are named differently on different OS's, running everything in a container simply removes this entire area of problems.

If you are okay with it, I'd like to start my first contribution on this project by writing the few configuration files required to containerize liberapay using Docker for a smooth development setup. I'm okay to write some documentation too regarding this alternative setup, and to explain more in depth why I think it could be a useful improvement, especially to welcome newcomers.

EliotBerriot:
This is a basic PoC of a development setup using Docker. **I've put the usage instructions in the README, but if you'd rather not have it here, we can easily move it somewhere else** (maybe there is a wiki or documentation you use internally).

Basically, if you want to start on the project and have Docker / docker-compose installed, you can now set up the whole thing in 3 commands:

    docker-compose build                        
    docker-compose run web bash recreate-schema.sh
    docker-compose run web python -m liberapay.utils.fake_data

Then run the application with:

    docker-compose up

The first one is the real Docker one, that spins up the application container (based on Debian 8, but this can be changed to reflect the production environment). The other ones are simple calls to existing setup scripts (database creation and population).

Tests with py.test are working in this setup, you can run the whole test suite with:

    docker-compose -f docker/tests.yml run tests

Note this accept all py.test arguments such as --ff (failed tests first) or -x (fail fast):

    docker-compose -f docker/tests.yml run tests --ff -x

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM debian:8
+
+RUN apt-get update && apt-get install build-essential wget libpq-dev python-dev postgresql-client -y
+
+RUN wget -O /tmp/get-pip.py https://bootstrap.pypa.io/get-pip.py && python /tmp/get-pip.py
+
+ADD ./requirements.txt /requirements/pip/base.txt
+RUN pip install -r /requirements/pip/base.txt
+
+ADD ./requirements_dev.txt /requirements/pip/dev.txt
+RUN pip install -r /requirements/pip/dev.txt
+
+ADD ./requirements_tests.txt /requirements/pip/tests.txt
+RUN pip install -r /requirements/pip/tests.txt
+
+ADD . /app
+WORKDIR /app
--- a/README.md
+++ b/README.md
@@ -136,6 +136,30 @@ To deploy the app simply run `release.sh`, it'll guide you through it. Of course
 
 [OS-dev]: https://developers.openshift.com/
 
+### Setting up a development environment using Docker
+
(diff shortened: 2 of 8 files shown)


Changaco:
It would be better to avoid duplicating the environment variables. You're installing the dev requirements so you have honcho inside the container, you can use it to launch the app the same way we do in the Makefile (probably with an additional env file for Docker-specific values).

EliotBerriot:
You're right, I'll look into that tonight

EliotBerriot:
Okay, I've update the Docker configuration to use honcho and source env file. The only problem is that honcho apparently does not proxy environment variables  that are not in an .env file, so there is now way to pass docker-only environment variables in docker-compose files.

I've tackled that by adding a `docker/defaults.env` and a `docker/tests.env` file with docker specific configuration.

Changaco:
Looks fine to me.

PR #451, opened 2016-12-01
Fix #132: Added explicit button to update nominal take in team table

### LINKED ISSUE #132: Add button "OK" on team management
On the page https://liberapay.com/about/me/receiving/#teams, when an user change his part, there are not button to say "ok, record that". 

"Enter key" work perfectly, but add a button too (with the same effect) is better for ergonomy.
# 

Je me permet en français, car je ne suis pas sûre d'être compréhensible... Sur la page https://liberapay.com/about/me/receiving/#teams , il manque l'indice visuel d'un bouton pour valider le fait qu'on vient de changer sa part dans la répartition du revenu. La touche "entrée" marche très bien, mais tout le monde n'est pas habitué à naviguer au clavier et ajouter un bouton "Valider" améliorera l'ergonomie.

EliotBerriot:
This was a simple one (as promised):

screenshot from 2016-12-01 17-41-14

However, on the long term, I think the UX would be better if the line relavent to the current user was put somewhere in a deterministic way, for example at the top, or to move the form outside the table. I imagine if you have a team with 10s of members, it will become painful to find yourself in that. Also, if you're not familiar with the UI, it can be easy to miss the input.

Also, because of how bootstrap handles input groups, the line holding the input will be taller than other ones.

### DIFF
--- a/templates/team-members.html
+++ b/templates/team-members.html
@@ -40,7 +40,10 @@
                     <form action="/~{{ team.id }}/income/take?back_to={{ urlquote(request.line.uri) }}"
                           method="POST">
                         <input type="hidden" name="csrf_token" value="{{ csrf_token }}" />
-                        <input class="form-control" name="take" value="{{ format_decimal(member.nominal_take, '#,##0.00') }}" />
+                        <div class="input-group">
+                            <input class="form-control" name="take" value="{{ format_decimal(member.nominal_take, '#,##0.00') }}" />
+                            <span class="input-group-btn"><button class="btn btn-default" title="{{ _("Update my nominal take") }}" type="submit"><i class="fa fa-edit"></i></button></span>
+                        </div>
                     </form>
                 </td>
             % else


Changaco:
> However, on the long term, I think the UX would be better if the line relavent to the current user was put somewhere in a deterministic way, for example at the top, or to move the form outside the table. I imagine if you have a team with 10s of members, it will become painful to find yourself in that. Also, if you're not familiar with the UI, it can be easy to miss the input.

The rows are sorted by username, so it's not that hard to find yours, and we could highlight it too, but I'm okay with moving the input outside the table if you want (in this case we should leave a little edit button where the input used to be, targeting the new input).

> Also, because of how bootstrap handles input groups, the line holding the input will be taller than other ones.

We have custom CSS rules for inputs inside a table cell, `.table td > form > input.form-control`, if you leave the input inside the cell then you should modify the selectors to accommodate the extra `.input-group` between the form and the input.

Changaco:
> if you leave the input inside the cell then you should modify the selectors

Actually, let's leave things as they are.

Changaco:
!m @EliotBerriot

Issue #277, opened 2016-04-29
Investigate implementing smoothState.js


Calinou:
[smoothState.js](https://github.com/miguel-perez/smoothState.js) is a jQuery plugin that makes page transitions feel faster. Optional animations are supported, and the library is based on the progressive enhancement philosophy - users with JavaScript disabled or with an unsupported browser can continue to browse the site just fine.

Changaco:
There might be some issues as we treat some AJAX requests differently than normal requests.

Changaco:
Same thing as a standalone instead of a jQuery plugin: https://instant.page/

Changaco:
JavaScript isn't really needed to do this anymore, as Chrome has added a mechanism to tell the browser to this itself by adding a `<script type="speculationrules">` tag in the HTML or a [`Speculation-Rules`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Speculation-Rules) HTTP header containing.

I learned about this from Cloudflare's blog post [Introducing Speed Brain: helping web pages load 45% faster](https://blog.cloudflare.com/introducing-speed-brain/). (I turned off Speed Brain for `liberapay.com`, it's useless since the Liberapay webapp doesn't have static HTML pages.)

Direct link to Google's documentation: [Prerender pages in Chrome for instant page navigations](https://developer.chrome.com/docs/web-platform/prerender-pages).

PR #453, opened 2016-12-01
Fix #136: display notifications directly in the nav bar via a popup

### LINKED ISSUE #136: Display notifications directly under navbar icon
We should add a little JavaScript to show the notifications as a kind of popup under the notifications icon, so you don't actually have to go the notifications page.

EliotBerriot:
### Remaining TODO (added by @Changaco)

- [ ] don't overflow on small screens
- [ ] only fetch the notifications when the user clicks on the button
- [ ] use notification titles instead of bodies (titles weren't available when this PR was started)

### Original message

This is a big PR, mainly because it involves javascript / API calls / frontend work to fix #136.

This also include a fix to the following race condition: if a user browses to the notification page and clicks on `mark as read` button and another notification was added during that time, this unread notification was previously marked as read too. This fix was done in the first commit of the PR, and basically involves passing the ID of the last notification you want to mark as read when calling `mark_notifications_as_read`.

This is my first contact with the database, so feel free to report anything that looks bad in the way I implemented that.

Another minor fix: on the notification page, the `mark as read` button is now disabled if there is now notification to mark as read. 

Now, for the big part: I tried to make that as light as possible, and to use only bootstrap component. I'm rather satistfied with the result:

notifications

This uses Bootstrap's Popover component, and only required a few lines of custom CSS. The JS is quite dirty, but I honestly think we can't implement easily that kind of behaviour using only JQuery. 

One of the tricky part was to actually remove the hard coded styles in the notification HTML. I understand it was designed for email notifications, so I added a flag in the `render_notifications` to achieve that when needed. Note that this old menu link should still work if javascript is disabled.

For the future, If we find out that we need a more interactive UI that involves more javascript, I suggest we use a micro framework to achieve that. I had success with [Vue.js](https://vuejs.org/) in the past for this kind of things. (if you're curious, I could take some time to rewrite this feature using Vue.js, just so you can see the benefits).

**I strongly advise someone more experienced than me on this project takes the time to test this**, I'd rather not break the front end on a silly mistake ;)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM debian:8
 
-RUN apt-get update && apt-get install build-essential wget libpq-dev python-dev postgresql-client -y
+RUN apt-get update && apt-get install build-essential wget libpq-dev python-dev postgresql-client libffi-dev libssl-dev -y
 
 RUN wget -O /tmp/get-pip.py https://bootstrap.pypa.io/get-pip.py && python /tmp/get-pip.py
 
--- a/js/10-base.js
+++ b/js/10-base.js
@@ -76,6 +76,8 @@ Liberapay.init = function() {
     });
 
     if (!!Liberapay.username){
+        // if the user is logged in, we initialize
+        // the notification popover
         Liberapay.setupNotifications();
     }
 };
--- a/js/10-base.js
+++ b/js/10-base.js
@@ -76,8 +76,7 @@ Liberapay.init = function() {
     });
 
     if (!!Liberapay.username){
-        // if the user is logged in, we initialize
-        // the notification popover
+        // if the user is logged in, we initialize the notification popover
         Liberapay.setupNotifications();
     }
 };
--- a/js/10-base.js
+++ b/js/10-base.js
@@ -74,6 +74,10 @@ Liberapay.init = function() {
     $('.radio-group .list-group-item > label').on('click', function() {
         $(this).children('input[type="radio"]').prop('checked', true);
     });
+
+    if (!!Liberapay.username){
+        Liberapay.setupNotifications();
+    }
 };
 
 Liberapay.error = function(jqXHR, textStatus, errorThrown) {
--- a/js/notification.js
+++ b/js/notification.js
@@ -23,3 +23,55 @@ Liberapay.notification = function(text, type, timeout) {
     $dialog.append($('<span class="close">&times;</span>').click(close));
     if (timeout > 0) setTimeout(close, timeout);
 };
+
+Liberapay.setupNotifications = function() {
+    // fetch notifications from the backend and setup the corresponding
+    // popup in the navbar
+
+    var url = '/' + Liberapay.username + '/notifications.json';
+    var link = $("nav .notifs");
+    var limit = 5;
+    $.get(url, function(data){
+        if (data.length === 0){
+            return;
+        }
+        link.on('click', function(e){
+            e.preventDefault();
+        });
+        link.popover({
+            html : true,
+            placement: 'bottom',
+            content: function() {
+                var content = $('.notifs-wrapper .data').html();
+
+                content += '<ul class="list-group">';
+                var notifs = data.slice(0, limit);
+                notifs.forEach(function(e){
+                    content += '<li class="list-group-item text-muted';
+                    if (e.is_new){
+                        content += ' unread';
+                    }
+                    content += '">';
+                    content += e.html;
+                    content += '</li>';
+                });
+                content += '</ul>';
+                $(document).on('click', 'nav .notifs-wrapper .popover .mark-read', function(e){
+                    var url = '/' + Liberapay.username + '/notifications.json'
+                    var data = {
+                        "mark_all_as_read": true,
+                        "until": notifs[0].id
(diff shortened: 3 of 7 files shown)


Changaco:
This is quite good. :-)

I've added a few commits to clean up the code.

The remaining issues are:

- [ ] clicking outside of the popover doesn't close it, and it doesn't have a close button either
- [ ] the navbar item isn't updated to reflect the new number of unread notifications, Vue.js could help with that I guess, but we can do without too

> This is a big PR

Nah, I'd say ~150 lines of diff is a medium size PR at most.

> I'd rather not break the front end on a silly mistake ;)

We don't have protections against everything (see the [Bug Prevention label](https://github.com/liberapay/liberapay.com/labels/Bug%20Prevention)), but overall it's not easy to break things without noticing it. ;-)

Changaco:
For future PRs it would be better to push to a branch in the main repo instead of in your fork, it makes it easier for others to contribute.

Changaco:
Oh yes, another thing: we're loading notifications on each page load instead of only when the user actually clicks on the navbar item, that's not very good.

EliotBerriot:
Okay, I'm trying to check the last boxes, but to be honest, this start looking like spaghetti code if we combine everything together: 

- If we initialize the whole think on the first click, there is a delay (depending on the network trip to fetch the notifications) before the popover display. Also, because the popover is not instanciated on document load directly but in an AJAX callback instead, some weird things are starting to happen
- on top of that, if you add the fact you should hide the popover on a click on the body, but you also need to keep the default event on the link of no notifications are available, it becomes really, really ugly.

Thi is the kind of code that will be likely to break at anytime and do unexpected things. If you are okay with that, I'll rewrite the whole thing using vue JS. I honestly think the whole thing will be easier to maintain, to understand and we can do much more with an adequate tool.

Changaco:
I am torn on the question of using Vue.js or not. I'm open to looking at a diff of the two possibilities. I would be more inclined to approve if you can show me that introducing Vue.js would enable us to simplify other parts of our existing JavaScript code and not just what we're building in this PR.

For the delay problem I suggest pre-loading using `$link.one('mouseover')` since that usually happens slightly before the click event. We have an issue about applying this technique to the whole website: #277.

EliotBerriot:
> I am torn on the question of using Vue.js or not. I'm open to looking at a diff of the two possibilities. I would be more inclined to approve if you can show me that introducing Vue.js would enable us to simplify other parts of our existing JavaScript code and not just what we're building in this PR.

I totally agree with that: ideally, a dependency should help with the whole project and not be used only in a single place.

I'll submit a patch with an example of the notification popover using VueJS, and I'll take a look at other javascript functionalities to see what could be refactored using VueJS (and if it would be really useful).

Changaco:
Ping @EliotBerriot. Happy new year! :-) Will you have some time to work on this in the coming weeks?

EliotBerriot:
Okay, this has been 6 months, it's time to close this. I'll try rebase my branch this week-end and submit another version using VueJS, so you can see the resulting code. 

I'm really sorry for the delay :/

EliotBerriot:
So here is the same branch, but including a rebase for the 6 last months, and minor commits : 

- Fixed docker setup that was broken
- Update the notification count when in the menu when notifications are marked as read
- Close the popover when you click on body

I'm not satisfied with PR as is though. It does not adress one of your concerns (loading the notifications when the popover link is clicked instead of on page load). And it's really ugly and barly maintenable since the javascript is really tied to the DOM structure.

I've suggested to use Vue some time ago to adress the DOM coupling part. However, Bootstrap javascript components (such as the Popover we use) are not made to display dynamically generated / changing content, and even with Vue, it becomes ugly (I've tried it). That kinds of defeat the benefits of using a dedicated framework such as Vue, because we have to store the popover HTML in a javascript variable, so DOM coupling again.

That leaves us a few choices:

- Live with it and deal with the dirty javascript / jquery code
- Reimplement a "dynamic-content-friendly" popover component to use with Vue (or another frontent framework)
- Abandon the idea of having a notification popover

@Changaco what's your mind on this?

Changaco:
To get a clearer view I have extracted and merged the race condition fix (#617), so that we're left with only the frontend code here.

In the current state I don't want to merge this, particularly because it's not responsive (it overflows on small screens).

I'm not convinced that we need a JS framework to implement this feature in an acceptable way. The JS doesn't have to be perfect (that's impossible), just good enough.

whit537:
Bump. Can we get a canonical todo list appended to the ticket description? It's unclear what the remaining work here is (apart from fixing merge conflicts ;).

Changaco:
> Can we get a canonical todo list appended to the ticket description?

Done.

whit537:
Worth starting a new PR for this with a local branch so we're not blocking on @EliotBerriot? Should be able to preserve his commits to date ...

EliotBerriot:
Please go for it and don't wait for me, it's been too long already :)

EliotBerriot:
(I'm closing but leaving the branch available if you want to reuse it)

mattbk:
Would :heart:  to see @chadwhitacre jump in with a PR, but I don't think this project is on his radar anymore.

Issue #177, opened 2016-02-22
Prevent downtime


Changaco:
The site was out for about an hour this morning, and it seems to be OpenShift's fault. :angry: :disappointed: 

This message appeared in the `rhc tail` I had running:

```
The system is going down for reboot in 1 minute!
This node will restart in 1 minutes. Please save your work. [Rebooting nodes via Ansible]
```

I see no error message indicating why the app wasn't restarted after the reboot.

`rhc app restart` brought the site back.

Changaco:
It looks like CloudFlare doesn't have alerts, they invite you to use [Pingdom](https://www.cloudflare.com/apps/pingdom/) instead, which is probably great for Pingdom's business, but sucks for us and our users. I really wish we had our own infrastructure.

Changaco:
The server rebooted again during the night. Fortunately this time I was running a monitoring script that restarted the app automatically. It was down for about 15 minutes according to the logs, which is still an unacceptably long time.

Changaco:
We've talked about this in IRC. I've sent an email to http://bearstech.com/ asking if they can provide us with an alternative to hosting on OpenShift.

Changaco:
[CloudFlare had some issues in Europe today](https://www.cloudflarestatus.com/incidents/j4r4c33w6b0f), at least one of our users was affected, he got a timeout error page and reported the problem [on Twitter](https://twitter.com/matunixe/status/744894887221407744).

matunixe:
Hello I'm the whiner from Twitter :-)
What about let go Cloudflare ? It's a very painful-in-the-ass product, block Tor, don't work in EU...

Angristan:
Why are you using openshift ? It seems pretty expensive.

Angristan:
@matunixe 

> What about let go Cloudflare ? It's a very painful-in-the-ass product, block Tor, don't work in EU...

Work perfectly in EU, it has only has issues today.

Changaco:
> What about let go Cloudflare ?

What would we replace it with? There are reasons why we use CloudFlare, and for the most part it does the job well.

As I've said before I wish we had our own infrastructure, but that costs time and money, and we're short on both.

> It's a very painful-in-the-ass product, block Tor, don't work in EU...

We've already white-listed Tor. As for today's connectivity issues, they could have happened to any Internet service provider.

> Why are you using openshift ? It seems pretty expensive.

[...]

PR #462, opened 2016-12-21
Automatically (re)start postgres in production

Changaco:
This is an attempt to address the problem described in <https://github.com/liberapay/liberapay.com/issues/177#issuecomment-264641041>. It's a bit weird to start a database server from a web app, but not necessarily a bad idea, we'll see if it works in practice.


Changaco:
It didn't work. /o\

PR #465, opened 2016-12-23
Secure `_blank` links

Changaco:
See <https://github.com/gratipay/gratipay.com/pull/4243>. (!m @EdOverflow)


EdOverflow:
I can confirm that all `target="_blank"` links are now secured against reverse tabnabbing. 👍

PR #467, opened 2016-12-26
Close open redirects to reduce phishing risks

Changaco:
Issue reported by @EdOverflow via email.


Changaco:
@EdOverflow You can test it now, I deployed it 20 minutes ago. ;-)

PR #471, opened 2016-12-27
Increase salt size to 32 bytes

EdOverflow:
Since we are using PBKDF2-SHA-256, the salt should be 32 bytes long.

For more inforamtion please refer to my recommendations here:
https://github.com/EdOverflow/hunter/blob/master/docs/password-storage.md

On a side note, how many rounds did you set?

### DIFF
--- a/liberapay/models/participant.py
+++ b/liberapay/models/participant.py
@@ -279,7 +279,7 @@ class Participant(Model, MixinTeam):
         if l < PASSWORD_MIN_SIZE or l > PASSWORD_MAX_SIZE:
             raise BadPasswordSize
         algo = 'sha256'
-        salt = urandom(21)
+        salt = urandom(32)
         rounds = website.app_conf.password_rounds
         hashed = cls._hash_password(password, algo, salt, rounds)
         hashed = '$'.join((
--- a/liberapay/models/participant.py
+++ b/liberapay/models/participant.py
@@ -285,6 +285,7 @@ class Participant(Model, MixinTeam):
             raise BadPasswordSize
         # Using SHA-256 as the HMAC algorithm (PBKDF2 + HMAC-SHA-256)
         algo = 'sha256'
+        # Generate 32 random bytes for the salt
         salt = urandom(32)
         rounds = website.app_conf.password_rounds
         hashed = cls._hash_password(password, algo, salt, rounds)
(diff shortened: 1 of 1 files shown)


Changaco:
password_rounds is 50000 in production.

EdOverflow:
Actually, would you mind if we started using the [passlib](https://passlib.readthedocs.io/en/stable/index.html) module? It has built-in features such as the ability to upgrade to a newer password hashing algorithm and [TOTP](https://passlib.readthedocs.io/en/stable/narr/totp-tutorial.html).

Changaco:
I would probably accept a pull request to use passlib, but it would actually increase security risks slightly, because we don't have integrity checking for dependencies, so even if we assume or determine that a specific version of passlib is secure we're not sure that the code we've vetted will be the one actually running in production.

EdOverflow:
OK, in that case I agree with you. It is better to continue using `hashlib`.

Changaco:
Actually, integrity checking (hashing) is [implemented in `pip` now](https://pip.pypa.io/en/stable/reference/pip_hash/) and available in our production environment, so let's set it up and then you can make a PR to use passlib if you want.

EdOverflow:
Great! I will get on to that as soon as possible.

Changaco:
I'm working on setting up pip hashing.

Changaco:
> I'm working on setting up pip hashing.

Done in #517.

Changaco:
ping @EdOverflow

EdOverflow:
ACK @Changaco.

Changaco:
@EdOverflow Will you be able to complete this PR or should someone else do it?

EdOverflow:
It would probably be easier if someone took over, but if you need any information or help, please feel free to ping me. 🙂

whit537:
Suggest closing as stale. Seems like moving to `passlib` should be a new PR.

Changaco:
Rebased on master and added the few lines of codes needed to update old password hashes.

Ready for review.

EdOverflow:
Looks good to me.

whit537:
Seems to deserve a test or two, since if we screw up the password rehash we risk locking people out of their account, no?

Changaco:
I had a feeling you were going to say that. I did a quick local trial but you're right that it would be better if the extra few lines I've added were covered by a test.

PR #476, opened 2016-12-30
History++

Changaco:
Closes #37, #254, and #396.

PR #479, opened 2017-01-01
History+

Changaco:
A few more improvements for the Wallet page. Previous: #476.


Changaco:
Before:

spectacle m27173

After:

spectacle h27173

PR #480, opened 2017-01-02
Handle Postgres downtime better

Changaco:
This PR fixes postgres autostart (#462 was flawed), but more importantly it modifies Liberapay to keep running and serving both static resources and proper error pages when the DB is down.

With this in place I think a few minutes of downtime to upgrade postgres will be acceptable, which will allow me to push back setting up zero-downtime upgrades to later.


Changaco:
Screenshot:

liberapay-2017-01-02-no-db

PR #488, opened 2017-01-04
Fix XSS on payment pages

Changaco:
This PR fixes a big security issue. The branch has already been pushed to production.

I've found no sign in the logs of attempts to exploit this vulnerability.

The issue does not affect Gratipay so there was no need to coordinate with them.

It's thanks to @EdOverflow's warning about open redirects (#467) that I've discovered this today.

### DIFF
--- a/liberapay/main.py
+++ b/liberapay/main.py
@@ -161,17 +161,23 @@ def _success(self, code=200, msg=''):
     raise self
 pando.Response.success = _success
 
+if hasattr(pando.Response, 'sanitize_untrusted_url'):
+    raise Warning('pando.Response.sanitize_untrusted_url() already exists')
+def _sanitize_untrusted_url(response, url):
+    if isinstance(url, bytes):
+        url = url.decode('utf8', 'replace')
+    if not url.startswith('/') or url.startswith('//'):
+        url = '/?bad_redirect=' + urlquote(url)
+    host = response.request.headers[b'Host'].decode('ascii')
+    # ^ this is safe because we don't accept requests with unknown hosts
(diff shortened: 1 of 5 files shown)


EdOverflow:
That slipped under my radar. Were you able to make the XSS payload fire?

Changaco:
Yes, the payload executed when clicking the vulnerable links ("Go back" and "Cancel" on payin and payout pages).

EdOverflow:
Would you mind giving me a PoC, because I am unable to reproduce this issue?

Changaco:
Possible wide solutions:

- filter out `javascript:` links client-side and/or server-side
- remove `'unsafe-inline'` from `script-src` in our CSP
- sanitize all link targets automatically server-side, somehow
- stop storing redirect URLs in the path and querystring

Changaco:
How to reproduce:

- run version 153 of liberapay locally
- create an account and set up a donation (otherwise the app doesn't let you start a payin)
- go to http://localhost:8339/about/me/wallet/payin/bankwire/amF2YXNjcmlwdDphbGVydCgiZm9vYmFyIikK
- click the Cancel button (bottom-right corner)
- you get a JS alert with the message "foobar"

EdOverflow:
All the solutions listed above are good, but keep in mind that by removing `unsafe-inline` from `script-src` you are relying on the CSP. I would definitely ensure that all input is escaped server-side.

Changaco:
I've created separate issues for ideas on defending against XSS attacks, I'm merging this so I can move forward with other PRs.

PR #489, opened 2017-01-04
Prevent timing attacks

EdOverflow:
Liberapay is vulnerable to timing attacks, because the `==` operation does a byte-by-byte comparison of two values and as soon as the two differentiate it terminates.

For more on timing attacks please refer to: https://github.com/EdOverflow/hunter/blob/master/docs/timing-attacks.md

### DIFF
--- a/liberapay/models/participant.py
+++ b/liberapay/models/participant.py
@@ -194,7 +194,7 @@ class Participant(Model, MixinTeam):
             algo, rounds, salt, hashed = p.password.split('$', 3)
             rounds = int(rounds)
             salt, hashed = b64decode(salt), b64decode(hashed)
-            if cls._hash_password(v2, algo, salt, rounds) == hashed:
+            if constant_time_compare(cls._hash_password(v2, algo, salt, rounds), hashed):
                 p.authenticated = True
                 return p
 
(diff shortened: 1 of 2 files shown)


Changaco:
The risk of a successful timing attack on a hash seems pretty low to me, but I guess it doesn't hurt much to do a constant time comparison here as well. We have our own function: `liberapay.security.crypto.constant_time_compare`.

EdOverflow:
I agree that this is not a serious issue, but nevertheless it requires fixing. Would you rather I used  `liberapay.security.crypto.constant_time_compare`?

---

`CVSS:3.0/AV:L/AC:H/PR:N/UI:N/S:U/C:L/I:L/A:N` (CVSS score: 4.0)

Changaco:
Yes, I would prefer that we only use one function for constant time comparisons.

EdOverflow:
Why are you using `bytes != str` instead of `isinstance(val2, bytes)`?

Changaco:
`bytes != str` is a python version detection, it's `False` on python 2 and `True` on python 3.

Changaco:
It wouldn't help, the function will fail either way if the strings are of different type on python 3.

EdOverflow:
In that case, it all looks good to me.

PR #468, opened 2016-12-26
Add Content Security Policy (CSP)

EdOverflow:
Content Security Policy (CSP) is a security mechanism that allows web sites control how browsers process their pages. In essence, sites can restrict what types of resources are loaded and from where. CSP policies can be used to defend against cross-site scripting, prevent mixed content issues, as well as report violations for investigation.

PR #498, opened 2017-01-05
Switch Content Security Policy from report-only to enforceable

Changaco:
In #468 we opted to start with a report-only policy. This PR is so we don't forget to switch to enforcement mode. I want to wait a little while longer and analyze the reports before merging this.

### DIFF
--- a/liberapay/security/__init__.py
+++ b/liberapay/security/__init__.py
@@ -30,17 +30,17 @@ def set_default_security_headers(website, response, request=None):
     # https://scotthelme.co.uk/content-security-policy-an-introduction/
     if b'content-security-policy' not in response.headers:
         csp = (
-            b"default-src %(main_domain)s;"
-            b"script-src %(main_domain)s 'unsafe-inline';"
-            b"style-src %(main_domain)s 'unsafe-inline';"
+            b"default-src 'self' %(main_domain)s;"
+            b"script-src 'self' %(main_domain)s 'unsafe-inline';"
+            b"style-src 'self' %(main_domain)s 'unsafe-inline';"
             b"connect-src *;"  # for credit card data
             b"img-src *;"
             b"reflected-xss block;"
         ) % {b'main_domain': website.canonical_host.encode('ascii')}
         csp += website.env.csp_extra.encode()
         if website.canonical_scheme == 'https':
-            csp += b"upgrade-insecure-requests;block-all-mixed-content;"
-        response.headers[b'content-security-policy-report-only'] = csp
+            csp += b"upgrade-insecure-requests;"
+        response.headers[b'content-security-policy'] = csp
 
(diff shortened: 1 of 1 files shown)


Changaco:
From the CSP reports we've received so far it looks like we need to change `'self'` to `liberapay.com` (actually `website.env.canonical_host`) for the i18n subdomains to be able to load the assets from the top domain.

Changaco:
I'm seeing reports for `script-src` `eval`. [Mozilla's wiki](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#Unsafe_eval_expressions) says that in addition to the `eval()` function itself, four other things are blocked when `unsafe-eval` isn't allowed. Searching for them in our code I see that `Function()` is used in the old version of jQuery we're using, so I'm going to upgrade that.

Changaco:
I'm also seeing reports for `block-all-mixed-content`, due to images included in profile statements. Fixing that requires an image proxy (<https://github.com/liberapay/liberapay.com/issues/504>).

strugee:
@Changaco FWIW you can send `Content-Security-Policy` and `Content-Security-Policy-Report-Only` at the same time, so you should be able to set everything but `block-all-mixed-content` to enforcing and keep that one as reporting: https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP#Testing_your_policy

Changaco:
I've dropped `block-all-mixed-content` since we already use `upgrade-insecure-requests` anyway.

I've also added `'self'` back to the `*-src` directives just in case.

Changaco:
I'll merge this tomorrow unless there are objections. Ping @EdOverflow.

liberapay/security/__init__.py
```diff
@@ -30,17 +30,17 @@ def set_default_security_headers(website, response, request=None):
     # https://scotthelme.co.uk/content-security-policy-an-introduction/
     if b'content-security-policy' not in response.headers:
         csp = (
-            b"default-src %(main_domain)s;"
-            b"script-src %(main_domain)s 'unsafe-inline';"
-            b"style-src %(main_domain)s 'unsafe-inline';"
+            b"default-src 'self' %(main_domain)s;"
+            b"script-src 'self' %(main_domain)s 'unsafe-inline';"
```
EdOverflow: The comment above this section of code suggests that we are using this CSP to prevent XSS, but by using `'unsafe-inline'` you are doing precisely the opposite.

EdOverflow: Proof of concept: http://output.jsbin.com/segekehedo.

Changaco: We're not ready to drop `'unsafe-inline'`. It's still better to have some protection than none at all.

EdOverflow: @Changaco, I agree, but is there a ticket addressing this concern somewhere?

Changaco: There is now: #879.

PR #502, opened 2017-01-09
Upgrade Markdown lib and allow XMPP URIs

Changaco:
Closes #500.

@EdOverflow Can you spare some time to review this? Or maybe play around with it after it's deployed? I'd appreciate any reassurance that this new code isn't leaving the door open to some XSS attacks.

PR #503, opened 2017-01-10
Fix double escaping in Markdown renderer

Changaco:
#502


Changaco:
This is deployed, double escaping is fixed.

Issue #487, opened 2017-01-04
Can't easily see who recently joined


Changaco:
As demonstrated by https://github.com/liberapay/salon/issues/97#issuecomment-270340668 even I don't see everything (but I really should because it's part of fraud detection).

We don't want to go back to automatically showing everyone who joins, because we don't want to promote new users we don't like. So, admins (I'm still the only admin as of now) should have a dashboard showing everyone and having buttons to white-list users we're okay with promoting. Then we can bring back a listing of users who recently joined, but it'll be a curated list this time.

Changaco:
Closing in favor of #512.

PR #506, opened 2017-01-17
Better control referencing

Changaco:
This PR is the first step towards fixing #487.

[Our Terms of Service](https://liberapay.com/about/legal#terms) state:

>The service normally allows the user to increase the visibility of its profile in various way, but the organization reserves the right to disable the user's access to those features.

However blocking a user's access to "publicity" wasn't actually implemented. Once this PR is deployed all user profiles, as well as community and pledge pages, will have `noindex` and `nofollow` turned on by default, and admins will also be able to prevent pages from being listed on Liberapay (`hide_from_lists`) and appearing in search results (`hide_from_search`).

I'm tagging this PR with "Self-defense" because it's about preventing "naughty" people from abusing Liberapay.

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -16,7 +16,18 @@ Liberapay.forms.jsSubmit = function() {
     function submit(e) {
         e.preventDefault();
         var $form = $(this.form || this);
+        var target = $form.attr('action');
+        var js_only = target == 'javascript:';
         var data = $form.serializeArray();
+        if (js_only) {
+            // workaround for http://stackoverflow.com/q/11424037/2729778
+            $form.find('input[type="checkbox"]').each(function () {
+                var $input = $(this);
+                if (!$input.prop('checked')) {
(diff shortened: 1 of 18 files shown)


Changaco:
Related issue: <https://github.com/liberapay/liberapay.org/issues/15>.

Changaco:
This is in production, no errors during deployment.

Changaco:
However, since I chose to apply the schema changes after deployment, postgres.py was still treating the modified participant columns as booleans (this is a corner case that wasn't fixed by <https://github.com/gratipay/postgres.py/pull/43>). Restarting the workers fixed the problem (`rhc ssh liberapay 'cd $OPENSHIFT_REPO_DIR && ./.openshift/action_hooks/post_deploy'`).

Changaco:
I completely forgot to add a checkbox for `profile_nofollow` in the new admin dashboard, doh!

PR #515, opened 2017-02-02
Fix payday stats

### LINKED ISSUE #437: Payday stats are misleading
Refunds are counted as withdrawals, and the refunded payins aren't removed from deposits because payday stats are computed during payday and never corrected after.

Changaco:
Closes #437. I've already tested this locally on a backup of the production DB.


Changaco:
Stats are fixed. https://liberapay.com/about/stats

spectacle m12284

PR #517, opened 2017-02-06
Automatically check the integrity of dependencies

Changaco:
Background: <https://github.com/liberapay/liberapay.com/pull/471#issuecomment-269451400>.


Changaco:
I've already tested this in production, and I've reinstalled all dependencies to make sure there wasn't any tampering in the past: `rhc ssh liberapay 'pip install --require-hashes --upgrade --force-reinstall -r /dev/stdin' <requirements.txt`.

PR #537, opened 2017-02-25
Add LICENSE

revi:
This commit will make github shows license information in the github web interface (next to the contributors' list) and it is a standard way to store license.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,121 @@
+Creative Commons Legal Code
+
+CC0 1.0 Universal
+
+    CREATIVE COMMONS CORPORATION IS NOT A LAW FIRM AND DOES NOT PROVIDE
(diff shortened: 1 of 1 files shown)


Changaco:
I strongly dislike putting LICENSE files everywhere, it's a waste of space. In fact one of the things I like in the CC0 is that it has no stupid "you must distribute the license with all copies" clause.

revi:
Sorry for late reply, I just wanted to see "CC0" in github UI, but in that case it makes sense for you.

comradekingu:
@Changaco Copyleft licensing (what you describe) carries copyleft, not redundancy. You are confusing it with something like the naming clause in old BSD, requiring the name of all original authors to follow the project license recursivly.

PR #555, opened 2017-03-16
Upgrade mangopaysdk to 3.0.6

### LINKED ISSUE #554: birthday will be saved one day early
when i want to save 1981-09-05, after refresh its 1981-09-04
when i input 1981-09-06 its saves correctly 1981-09-05 
(google chrome Version 57.0.2987.98)

strange....

Changaco:
Should fix #554.


Changaco:
Deploying this broke the mangopay module completely, so I rolled back to version 172.

The error message isn't very helpful: `APIError: {"error":"invalid_client"}`.

Changaco:
Problem fixed in c7797eadecb9d1c88e16943589353e99c975920d.

Issue #553, opened 2017-03-13
Migration to AWS


Changaco:
I'm going to use this issue to log my attempt to migrate Liberapay to AWS. See #177 for background.

Changaco:
I've installed https://github.com/aws/aws-cli on my machine (`pacman -S aws-cli`).

Changaco:
I've also [installed the Elastic Beanstalk CLI](http://docs.aws.amazon.com/elasticbeanstalk/latest/dg/eb-cli3-install.html).

Changaco:
I've created an AWS account for Liberapay, it required: accepting the [AWS Customer Agreement](https://aws.amazon.com/agreement/), a postal address, a credit card number, a phone number (you receive a call and input a 4-digit code), and finally choosing a support plan.

Changaco:
Okay, I have the Liberapay app running on AWS, with an empty DB. The steps to get to that were:

- create an access key in the [IAM console](https://console.aws.amazon.com/iam/home)
- create the "liberapay.com" application: `eb init`
- create the "production" environment: `eb create`
- add [.ebextensions](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/ebextensions.html) to:
    - install `postgresql95-devel` (for `psycopg2`) and `libffi-devel` (for `misaka`)
    - install our python dependencies with an up-to-date version of `pip`
- set the `WSGIPath` to `liberapay/main.py` via `eb config`, and add an `application = website` alias in that file
- import the environment variables from OpenShift: `rhc env list | xargs -d '\n' eb setenv`
- adjust the environment variables: drop `GUNIORN_OPTS`, create `DATABASE_URL`
- implement variable interpolation for `DATABASE_URL` in `liberapay/wireup.py`
- run `psql <sql/recreate-schema.sql` and `psql <sql/app-conf-defaults.sql` via `eb ssh`

Changaco:
I've been familiarizing myself with AWS, here are some notes:

- DB encryption isn't available for micro instances, so we won't have it for now
- Elastic Beanstalk seems to use S3 a lot for its internal workings, eating our free tier allotment of S3 requests (here's a forum thread about it: [S3 used for Elastic Beanstalk without explicit user consent](https://forums.aws.amazon.com/thread.jspa?threadID=239773))
- S3 doesn't keep logs of requests by default, and they're only "best effort" if you turn them on (meaning that log messages can be dropped)
[...]

PR #556, opened 2017-03-17
Read-only mode

Changaco:
Part of the AWS migration plan: <https://github.com/liberapay/liberapay.com/issues/553#issuecomment-286675786>.


Changaco:
This worked fine in practice, but it requires closing all open connections to the DB, because the `default_transaction_read_only` setting seems to be cached at the connection level.

PR #558, opened 2017-03-22
Fix chart overflow

### LINKED ISSUE #552: Charts can overflow
Screenshot of [the Stats page](https://liberapay.com/about/stats):

spectacle lh6293

Changaco:
Closes #552.


Changaco:
Note that this is only a partial fix, the charts will still overflow on small screens, and on larger screens in less than a year.

PR #568, opened 2017-04-13
Typo correction in www/index.html.spt

jorgesumle:
The sentence is singular, so it shouldn't use the word 'then'. I didn't write "him" or "her" because we don't know the gender that person. So I rewrote the sentence a bit, so that it is gender friendly. What do you think?

### DIFF
--- a/www/index.html.spt
+++ b/www/index.html.spt
@@ -226,7 +226,7 @@ recent = query_cache.one("""
         ) }}</p>
         % if sponsors
             <p>{{ ngettext(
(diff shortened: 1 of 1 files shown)


Changaco:
It's not a typo, "them" is sometimes used in English as a singular non-gendered pronoun. https://en.wikipedia.org/wiki/Singular_they

jorgesumle:
Thanks for the link and the explanation. I didn't know that, sorry.

Issue #288, opened 2016-05-02
Profile pages lack metadata


Changaco:
See [Twitter Cards](https://dev.twitter.com/cards/overview) and similar docs of other social networks.

jorgesumle:
Maybe this is useful: http://ogp.me/

Issue #269, opened 2016-04-16
"X alternative" pages


Changaco:
We could create [alternative pages](http://www.gkogan.co/blog/alternative-pages/) to explain to users of other crowdfunding platforms why they might want to use Liberapay.

http://alternativeto.net/software/liberapay/ is already one of our top referrers, but that website is only in English, while our pages could be translated.

Related issue: #259.

julia-r:
Product Hunt also has a page with alternatives to Patreon, maybe Liberapay can be added there as well: https://www.producthunt.com/alternatives/patreon

Changaco:
It wouldn't be appropriate for me or someone else deeply involved in Liberapay to submit it on websites like Product Hunt. Ideally it should be done by someone who already has an account on Product Hunt and isn't part of the Liberapay team.

julia-r:
Made a Product Hunt page: https://www.producthunt.com/posts/liberapay ;)

Changaco:
Implementing this could help answer questions like <https://infosec.exchange/@hashtagsecurity/101352346179511165> and <https://octodon.social/@sciss/101528434919145207>.

mattbk:
Oh wow, I did not see all those other replies through my Mastodon app. :thinking:

Changaco:
Alternative pages would also be a good place to state clearly that Liberapay doesn't support rewards (#1428) and explain why.

Changaco:
I've opened an embryonic Pull Request for this: #1432.

unfa:
I think this is a good idea.

I myself listed a few differences between Patreon and Liberapay for my Patreon supporters:
https://www.patreon.com/posts/launched-on-24469484

I've finally decided to launch on Liberapay, as I found out about a recent Patreon controversy.

I think a page targeted at Patreon users might make me launch on Liberapay way earlier, as I've set up my account about a year ago, but only this week I have finalized and promoted it.

PR #571, opened 2017-04-15
WIP: add some HTML metadata to increase the visibility of website

jorgesumle:
This pull request aims to improving Liberapay's visibility in search engines by using HTML meta tags (documentation for HTML 5.1 meta tags: https://www.w3.org/TR/html51/document-metadata.html#the-meta-element). This pull request is a <abbr title="Work in progress">WIP</abbr> because it still lacks feedback and a discussion of possible improvements to the pull request.

The most important HTML metadata that is missing is the description, which is the first thing that users read when they see the abstract from a search result. According to HTML 5.1 W3C Recommendation (emphasis made by me): 

> The value must be a free-form string that describes the page. **The value must be appropriate for use in a directory of pages, e.g., in a search engine.** There must not be more than one meta element with its name attribute set to the value description per document.

As you can see this is very important because its the first impression people get when they see Liberapay's abstract in a search engine. It must have the appropriate length also, it mustn't be too short or too long.

I have looked at the description meta tag of a website that is doubtless expert in SEO, a search engine called DuckDuckGo. The description is exactly what is written in the meta tag. The length of its text string is 111 (19 words); the length of the description string in this pull request is 121 (19 words).

I don't think the keywords are very important to SEO, but it does not hurt to have them. Let's see what the HTML 5.1 W3C Recommendation says about them:

> The value must be a set of comma-separated tokens, each of which is a keyword relevant to the page.

Note that I'm not a SEO expert and the description or the keywords I chose may not be the best: we should discuss this and provide other descriptions or keywords, then provide arguments, and finally make a decision. I chose this description because I thought it is what best describes Liberapay. I don't know if all the people will understand what the commons is or if there is a better way to express that, but here are my 2 cents:

> Liberapay is a recurrent donations platform which aims to provide a stable income for people contributing to the commons.

Regarding the keywords, I chose them because it is what I think most people who are searching something like Liberapay in a search engine would type.

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -7,6 +7,8 @@
     {% block head_early %}{% endblock %}
     <title>{% if title %}{% if subhead %}{{ subhead }} - {% endif %}{{ title }} - {% endif %}Liberapay</title>
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta name="description" content="{{ _("Liberapay is a recurrent donations platform which aims to provide a stable income for people contributing to the commons.") }}">
+    <meta name="keywords" content="{{ _("commons,crowdfunding,donations,free culture,free software,funding,money,platform,recurrent") }}">
     <link rel="icon" href="{{ website.asset('liberapay/icon-v2_black.16.png') }}" type="image/png">
     <link rel="stylesheet" href="{{ website.asset('bootstrap/css/bootstrap.css') }}">
     <link rel="stylesheet" href="{{ website.asset('base.css') }}" type="text/css">


Changaco:
I'm not sure having the same description on every page would be good.

We currently have this on the homepage: `<meta name="description" content="{{ _('Liberapay is a recurrent donations platform.') }}" />`

Related issues: #269 and #288.

jorgesumle:
> I'm not sure having the same description on every page would be good.

Mmm... Yes, that doesn't seem wise. I missed that it was in the homepage.

Meta keywords seem to be lacking though. What do you think? Is it worth adding them?

Changaco:
> Meta keywords seem to be lacking though. What do you think? Is it worth adding them?

I'm not sure, I think search engines don't really care about them, but I'm not an expert in SEO.

PR #580, opened 2017-04-20
Add metadata to user pages

jorgesumle:
Tries to solve #288.

To do:

- [x] <del>Profile's first and last name</del>
- [ ] og:description. Not sure what to put there yet. Maybe the goal, or what the person is receiving and donating?
- [x] Add the http://ogp.me/ns/profile# namespace

### DIFF
--- a/templates/profile-base.html
+++ b/templates/profile-base.html
@@ -6,15 +6,11 @@
     % if nofollow or noindex
     <meta name="robots" content="{{ ','.join(noindex + nofollow) }}" />
     % endif
-    <meta name="description" content="{{ _("{0}'s profile in Liberapay", participant.username) }}">
-    <meta name="og:title" content="{{ participant.username }}">
-    <meta name="og:type" content="profile">
-    <meta name="og:url" content="{{ website.canonical_scheme + '://' + request.hostname + '/' + participant.username }}">
-    <meta name="og:image" content="{{ participant.avatar_url }}">
-    <meta name="og:description" content="???">
-    % for l_completion, l_code, l_name, l_url in website.lang_list
-        <meta name="og:locale:alternate" content="{{ l_code }}">
-    % endfor
-    <meta name="og:site_name" content="Liberapay">
-    <meta name="profile:username" content="{{ participant.username }}">
+    <meta property="og:title" content="{{ full_title }}">
+    <meta property="og:type" content="profile">
+    <meta property="og:url" content="{{ website.canonical_scheme + '://' + request.hostname + '/' + participant.username }}">
+    <meta property="og:image" content="{{ participant.avatar_url }}">
+    <meta property="og:site_name" content="Liberapay">
+    <meta name="twitter:card" content="summary">
+    <meta name="twitter:site" content="@liberapaye">
 % endblock
--- a/templates/profile-base.html
+++ b/templates/profile-base.html
@@ -6,4 +6,15 @@
     % if nofollow or noindex
     <meta name="robots" content="{{ ','.join(noindex + nofollow) }}" />
     % endif
+    <meta name="description" content="{{ _("{0}'s profile in Liberapay", participant.username) }}">
+    <meta name="og:title" content="{{ participant.username }}">
+    <meta name="og:type" content="profile">
+    <meta name="og:url" content="{{ website.canonical_scheme + '://' + request.hostname + '/' + participant.username }}">
+    <meta name="og:image" content="{{ participant.avatar_url }}">
+    <meta name="og:description" content="???">
+    % for l_completion, l_code, l_name, l_url in website.lang_list
+        <meta name="og:locale:alternate" content="{{ l_code }}">
+    % endfor
+    <meta name="og:site_name" content="Liberapay">
+    <meta name="profile:username" content="{{ participant.username }}">
 % endblock
--- a/liberapay/exceptions.py
+++ b/liberapay/exceptions.py
@@ -205,7 +205,8 @@ class BadAmount(LazyResponse400):
         period = self.args[1]
         return _(BAD_AMOUNT_MESSAGES[period], self.args[0], *DONATION_LIMITS[period])
 
-class UserDoesntAcceptTips(LazyResponse400):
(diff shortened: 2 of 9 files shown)


templates/profile-base.html
```diff
@@ -4,6 +4,20 @@
...
+    <meta name="og:url" content="{{ website.canonical_scheme + '://' + request.hostname + '/' + participant.username }}">
+    <meta name="og:image" content="{{ participant.avatar_url }}">
+    <meta name="og:description" content="???">
+    <meta name="og:locale" content="{{ locale.language }}">
+    % for l_completion, l_code, l_name, l_url in website.lang_list
+        <meta name="og:locale:alternate" content="{{ l_code }}">
+    % endfor
+    <meta name="og:site_name" content="Liberapay">
+    <meta name="profile:first_name" content="">
+    <meta name="profile:last_name" content="???">
```
Changaco: We don't store the birth names of users in our database, and we shouldn't display them publicly without consent.

templates/profile-base.html
```diff
@@ -4,6 +4,20 @@
...
-    <meta name="robots" content="{{ ','.join(noindex + nofollow) }}" />
+    <meta name="robots" content="{{ ','.join(noindex + nofollow) }}">
     % endif
+    <meta name="description" content="{{ _("%s's profile in Liberapay" % participant.username ) }}">
+    <meta name="og:title" content="{{ participant.username }}">
+    <meta name="og:type" content="profile">
+    <meta name="og:url" content="{{ website.canonical_scheme + '://' + request.hostname + '/' + participant.username }}">
+    <meta name="og:image" content="{{ participant.avatar_url }}">
+    <meta name="og:description" content="???">
+    <meta name="og:locale" content="{{ locale.language }}">
```
Changaco: `locale.language` is only constant in subdomains (e.g. <https://en.liberapay.com/>), on <https://liberapay.com/> it depends on the request header `Accept-Language`.

jorgesumle:
OK, I applied some of your corrections.

The only thing left is the `og:description`:
> og:description - A one to two sentence description of your object.

This is important, because it's the text that people will see when they share your profile. What should we display there? I think that it would be good to display the first two sentences of the statement, because that is what will engage and inform people about your objective, contributions, etc. The sentences must be delimited by '.', '?', '!'... (am I missing another language that uses another punctuation?) and, of course, must correspond to the language of the link. If the link is https://es.liberapay.com/username/ it must pick the Spanish description if available, else it should use another language.

We could also display the money goal. There are some people that don't have a statement, so for them we could display what they give, earn, their objectives...

Any thoughts?

Changaco:
Extracting the first two sentences of a text is quite complex. I think we should ask users to input a short description along with their full profile statement.

I'm not sure we want the same description for the profile page (e.g. <https://liberapay.com/Liberapay/>) and the donation page (e.g. <https://liberapay.com/Liberapay/donate>). So it shouldn't be set in `templates/profile-base.html` but in `www/%username/index.html.spt` and `www/%username/donate.spt`.

templates/profile-base.html
```diff
@@ -4,6 +4,17 @@
...
     % endif
+    <meta name="description" content="{{ _("{0}'s profile in Liberapay", participant.username) }}">
+    <meta name="og:title" content="{{ participant.username }}">
+    <meta name="og:type" content="profile">
+    <meta name="og:url" content="{{ website.canonical_scheme + '://' + request.hostname + '/' + participant.username }}">
+    <meta name="og:image" content="{{ participant.avatar_url }}">
+    <meta name="og:description" content="???">
+    % for l_completion, l_code, l_name, l_url in website.lang_list
+        <meta name="og:locale:alternate" content="{{ l_code }}">
+    % endfor
```
Changaco: It's weird to have `og:locale:alternate` but no `og:locale`.

jorgesumle:
> Extracting the first two sentences of a text is quite complex. I think we should ask users to input a short description along with their full profile statement.

That would work. Also, for that short description we should set a reasonable character limit. We could display that description in the Explore>Individuals page or somewhere else too. Why? Maybe in the organizations page is more obvious to find out what an organization does without reading the statement, but in the individuals page the only thing I can see is a lot of people, who I don't know what they do who want to accomplish a goal or are receiving money. A short description would be useful there.

I find it overkill to make users introduce a short statement only for `og:description`.

Changaco:
Ready for review.

Changaco:
ping @jorgesumle

jorgesumle:
Looks good for me, but I haven't tested it. It should work fine on Diaspora* and other social networks. I was playing with Jinja2 to see if we could avoid leaving a word incomplete (…), I could do that, but I didn't like the code very much.
```
{% set og_description = og_description[:SUMMARY_LENGTH].split(' ')[:-1]|join(' ') %}
{% if og_description[-1] in ['.', '?', ':', '!'] %}
  {% set og_description = og_description[:-1] + '…' %}
{% endif %}
```
 It's very complicated, slower and I'm surely missing some language. I prefer your code for the function `excerpt_intro`.

This is a nice feature for users who have social networks. After it gets merged, it would be nice to write a small update.

Changaco:
Preview in [Twitter's Card Validator](https://cards-dev.twitter.com/validator):

screenshot-2017-10-4 card validator twitter developers

---

screenshot-2017-10-4 card validator twitter developers 1

Notes on this second card:

- The first attempt failed because `og:title` was empty, I fixed that in ab85b1ccb4f94136b3d35647acb2387509315162.
- My avatar isn't shown, I'm not sure why.

jorgesumle:
The profile avatar is shown in Diaspora* in my case. The URL is correct. Does Twitter require the URL to be from the same domain?

captura de pantalla de 2017-10-04 15-52-36

Changaco:
I doubt they have a same-domain restriction, many websites serve images from separate domains.

Maybe the validator doesn't actually try to fetch new images.

Changaco:
@jorgesumle Note that if you use the subdomain that corresponds to the language of your post, i.e. https://es.liberapay.com/jorgesumle instead of <https://liberapay.com/jorgesumle>, you'll get the summary in Spanish instead of English.

PR #583, opened 2017-04-28
Handle bankwire failures properly

### LINKED ISSUE #578: Callbacks for bankwire payin failures result in 500s
Two months ago MangoPay released an automatic expiration for old payin bankwires (<https://github.com/liberapay/salon/issues/98#issuecomment-280304403>). Last night their system sent us a callback for every expired bankwire payin we have.

`www/callbacks/mangopay.spt` needs to be modified to handle this type of event properly. That means recording the failure but not sending an email for old failures.

Changaco:
Closes #578.


Changaco:
I've grabbed the URLs for the callbacks that failed from the server logs, and replayed them (`while read url; do curl "https://liberapay.com$url"; sleep 1; done <callbacks.txt`).

PR #584, opened 2017-04-28
Compare email addresses in a case-insensitive way

### LINKED ISSUE #570: Email field is case sensitive.
Hello, I just tried to log into my account, and It seem that putting an caps inside one letter is actually considered as another email (on phone, the first letter is a lot of time caps'ed so it's silly to prevent it)
I think this should not be case sensitive, since emails aren't case sensitive in general :)

Steps to reproduce : 
 - Register
 - Log out
 - Try to log-in using different caps inside the email field from the one you registered.

Changaco:
Closes #570.

### DIFF
--- a/liberapay/models/participant.py
+++ b/liberapay/models/participant.py
@@ -151,16 +151,16 @@ class Participant(Model, MixinTeam):
 
     @classmethod
     def _from_thing(cls, thing, value):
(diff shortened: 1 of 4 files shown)


Changaco:
There was one user who had two accounts with the same lowercase email address, they're both closed, I modified one of them.

Issue #514, opened 2017-01-31
The UI isn't fully responsive


Changaco:
While #513 improves the responsiveness of several elements, others remain problematic, for example the tables on the Giving and Wallet pages.

PR #609, opened 2017-06-08
Bigger footer

### LINKED ISSUE #293: The footer is too small
The current format is low-visibility, and the number of links we can put in it is limited.

Bat41:
There is now two columns in the footer, so we can add as many links as we want. The first one is for internal links, the second one for social network accounts (I added a link to the Mastodon account by the way).

capture d ecran du 2017-06-08 17 58 52

Fix #293

### DIFF
--- a/style/base/base.scss
+++ b/style/base/base.scss
@@ -11,12 +11,12 @@ body {
         margin: 100px auto 20px auto;
 
         ul {
-          list-style: none;
-          padding: 0px;
+            list-style: none;
+            padding: 0px;
 
-          li {
-            margin: 10px auto;
-          }
+            li {
+                margin: 10px auto;
+            }
         }
     }
 }
--- a/style/base/base.scss
+++ b/style/base/base.scss
@@ -15,12 +15,16 @@ body {
             text-align: center;
         }
 
-        ul {
+        & > ul {
             list-style: none;
             padding: 0px;
 
-            li {
-                margin: 10px auto;
+            & > li {
+                display: inline-block;
+                & > a {
+                    display: inline-block;
+                    padding: 10px;
+                }
             }
         }
     }
--- a/style/base/base.scss
+++ b/style/base/base.scss
@@ -6,11 +6,17 @@ html, body {
 body {
     overflow-y: scroll;
     & > footer {
-        margin: 100px 0 20px 0;
-        text-align: center;
-        & > a {
-            display: inline-block;
-            padding: 10px;
+        display: flex;
+        justify-content: space-around;
+        margin: 100px auto 20px auto;
+
+        ul {
+          list-style: none;
+          padding: 0px;
+
+          li {
+            margin: 10px auto;
+          }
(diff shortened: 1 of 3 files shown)


style/base/base.scss
```diff
@@ -6,11 +6,17 @@ html, body {
...
+        justify-content: space-around;
+        margin: 100px auto 20px auto;
+
+        ul {
+          list-style: none;
+          padding: 0px;
+
+          li {
+            margin: 10px auto;
+          }
```
Changaco: You have inconsistent indentation here, 2 spaces instead of 4 in the rest of the file.

templates/base.html
```diff
@@ -112,15 +112,20 @@ <h1 class="{{ 'hidden-xs hidden-sm' if subhead }}">{{ title }}</h1>
...
+          <li><a href="/about/contact">{{ _("Contact Us") }}</a></li>
+          <li><a href="/about/faq">{{ _("FAQ") }}</a></li>
+          <li><a href="/about/legal">{{ _("Legal") }}</a></li>
+        </ul>
+        <ul>
+          <li><a href="https://mastodon.rocks/@Liberapay">Mastodon</a></li>
+          <li><a href="https://framasphere.org/u/liberapay">Diaspora*</a></li>
+          <li><a href="https://github.com/liberapay/liberapay.com">GitHub</a></li>
+          <li><a href="https://twitter.com/liberapaye">Twitter</a></li>
+          <li><a href="https://www.facebook.com/Liberapay-472128142945598/">Facebook</a></li>
```
Changaco: Same indentation inconsistency here.

Changaco:
I'm not familiar with the CSS flex model, it seems to be [okay for production use](http://caniuse.com/flexbox), I assume the two columns end up stacked on small screens?

Bat41:
Actually, the two columns stay as they are, but there isn't space anymore between them. But if you want I can make them stack.

Changaco:
The purpose of my question is to ensure that we don't worsen #514 by introducing elements that don't scale down well to small screen sizes. Do the columns overflow if you reduce the screen width to a ridiculous number of pixels, or do they wrap when there is no longer enough space?

Bat41:
No, they don't overflow. It's how it looks like on a 240 pixels large screen (the smallest size I can test with the dev tools of Firefox).

screen shot 2017-06-08 at 18 52 28

Changaco:
Okay, but that's only in English, the first column actually has a variable width if you take translations into account.

Bat41:
The column now wraps on small screens.

Changaco:
Screenshots:

screenshot_20170609_130743

---

screen shot 2017-06-09 at 13 06 41

---

screen shot 2017-06-09 at 17 03 38

Bat41:
That's nice! If there is no other problem, I think we can merge.

Changaco:
[!m](http://motivate.im/) @Bat41 💃

PR #615, opened 2017-06-16
Add support for repositories

Changaco:
This branch implements importing the list of a user's repositories from their GitHub or GitLab(.com) account. That allows the user to select repositories to show on their Liberapay profile. Repositories also show up in search results (even the ones that aren't shown on the profile page). The data about repositories is re-fetched regularly from the platforms, so stuff like the number of stars and the time of the last change don't become completely obsolete.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ matrix:
     - python: 3.6
       env: TOXENV=py36
 addons:
-  postgresql: 9.4
+  postgresql: 9.6
 branches:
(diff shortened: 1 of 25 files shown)


Changaco:
liberapay-2017-06-16-import-repos-changaco-gitlab

liberapay-2017-06-16-import-repos-liberapay-github

Changaco:
liberapay-2017-06-17-search-result-repo

PR #636, opened 2017-07-24
Reorganize "billing" module

Changaco:
A minor code cleanup.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,8 @@ branches:
 cache:
   directories:
     - .tox
+before_install:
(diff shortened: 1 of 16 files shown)


Changaco:
I've finally found a way to bypass Git LFS on Travis, by skipping Git entirely! In https://github.com/travis-ci/travis-yaml/blob/master/SPEC.md I discovered a setting named `git.strategy` that allows downloading a tarball instead of using `git clone`.

Changaco:
Travis is green \o/

PR #645, opened 2017-07-31
Implement switching account type

Changaco:
Closes #633 and #539.

PR #651, opened 2017-08-02
Fix wallet creation in `transfer()`

Changaco:
Fixes a regression from #645.


Changaco:
Added a fix for another bug I hit during <https://github.com/liberapay/salon/issues/152>.

PR #656, opened 2017-08-03
Thank our sponsors

### LINKED ISSUE #484: Sentry isn't mentioned anywhere
[Sentry](https://sentry.io/) is an essential part of keeping Liberapay as bug-free as possible. Now that they're sponsoring our use of their hosted service we really need to thank them properly, ideally we should have been doing that even when they weren't directly sponsoring us.

While we're at it we should also mention our use of Sentry's service in <https://liberapay.com/about/legal>.

Changaco:
Closes #484.


Changaco:
This PR adds a "Thanks" section to the homepage. It looks like this:

spectacle m20930

Changaco:
Any objections?

PR #666, opened 2017-08-13
Allow translating "Payday"

Changaco:
@pskosinski asked about this in IRC yesterday. I think translations should probably keep the name "payday" as-is, but adding a literal translation next to it could be useful.

### DIFF
--- a/www/about/money.spt
+++ b/www/about/money.spt
@@ -38,7 +38,7 @@ title = _("Money")
     "wallet per user, and only one currency (the euro) in that wallet."
 ) }}</p>
 
-<h3>Payday</h3>
+<h3>{{ _("Payday") }}</h3>
 
 <p>{{ _(
     "Payday is when donations are actually executed. It's a program ({0}this "


pskosinski:
On Weblate most translators converted "Last Payday" to their language and the only two translations keeping the "payday" word are Esperanto and Polish. I like your suggestion about adding a literal translation next to "payday". I think we could do that in all other strings containing "payday" too.

In Polish translation, for now, I was using genitive of payday: "payday'a". I worry it sounds very unnatural and I'd like to hear opinions of other native Polish speakers. I could replace "payday" with "payday process", so I can use genitive of "process" (just "proces" in Polish) and keep "payday" in nominative. The other option is using literal translation and adding the original in parentheses, like: "(…) to the level of the median take in the previous payday" -> "(…) do poziomu mediany poboru z ostatniego Dnia Wypłaty (Payday)", where "Dzień Wypłaty" is simply corresponding term in Polish and "Dnia Wypłaty" is naturally-sounding genitive.

I prefer the last option but with "Payday" header in /about/money containing only English word it didn't feel right, full. Thank you for "i18ning" it. 👍

tuxayo:
> the only two translations keeping the "payday" word are **Esperanto** and Polish.

And it's a matter of weeks before I remove them from the Esperanto translation. (except if we choose to keep Payday before the translation)

Issue #572, opened 2017-04-15
Donate button is not internationalized when JavaScript is disabled


jorgesumle:
The text should be the same when JavaScript is disabled:

![captura de pantalla de 2017-04-15 03 06 43](https://cloud.githubusercontent.com/assets/14212780/25059568/1c72d4ea-2189-11e7-93ab-60ded20270d5.png)

Changaco:
Generating SVGs with dynamic text is tricky, we would need to compute the correct button width for the translated word. Even then, the translations would get mixed up because of #360.

PR #670, opened 2017-08-19
Make donation button accessible

jorgesumle:


Changaco:
Looks okay to me, although I'm not sure it makes much sense to translate the `alt` attribute on an English-only image.

jorgesumle:
When https://github.com/liberapay/liberapay.com/issues/572 it fixed, it should be translated.

Changaco:
Is `role="button"` really appropriate and useful for a donation link? It may look like a button, but it's just a link to a Liberapay donation page. The [ARIA spec](https://www.w3.org/TR/wai-aria/roles) has a `link` role, but since we're already using a native link (`<a>`) we don't need to add the `role` attribute.

Issue #608, opened 2017-06-08
Design improvement


Bat41:
I would like to contribute to Liberapay. I don't know Python very well, I would like to help on the front-end. I started a little redesign of the team page.

![screen shot 2017-06-07 at 23 09 46-fullpage](https://user-images.githubusercontent.com/16254623/26931373-42d14b04-4c58-11e7-8c54-1346b78068a5.png)

I would like to have your opinion on this, and see how I could contribute (even if it's not by making style changes, I saw there is many "UI" issues I could work on).

Thanks for your feedback!

Changaco:
@Bat41 Welcome, and thank you.

The process of setting up a local development environment to be able to contribute to Liberapay's code is described in the README, if something doesn't work don't hesitate to ask for advice.

Maybe you could start with a small Pull Request to improve the footer (#293), which I see that you did in your screenshot.

I have some remarks about your proposed revamp of the profile pages:

- maybe the number of supporters/backers/donors should be in the same sentence as the income, e.g. "Mastodon receives €XX.XX per week from YY donors."
- perhaps the "joined X ago" should be in the "Income History" section
- is the Donate button visible enough or we should we keep the "glow" it currently has?

Bat41:
Thanks for your feedback! I completely agree with your two first remarks. For the third one, I personally don't like the « glowing » effect the button currently have . But maybe a subtle grey shadow could make it more visible…

I'll start by improving the footer.

Grummfy:
I'm not totally fan of rounded avatar (I have a squared one), but perhaps rounded corner like on github can be far nicer than the actuals.

Other wise I find it cleaner and more pleasant

lostinlight:
Hi, everyone.

I created a [design concept](https://goblinrefuge.com/mediagoblin/u/foggymorning/m/liberapay-design-exploration) for Liberapay's landing page. 

It's a far cry from current original version. Frankly, I wanted to work with given colours (like yellow in the logo), but it somehow didn't work out. So even if it remains just a design exploration, please, let me share some decisions I made, perhaps, they might come in handy.

1. I went for vibrant violet and greenish colours. The analogy of something fresh and liberating, like a gulp of mountain wind or the limitless deep ocean. Fonts are PT Sans Bold and Regular.

[...]

PR #672, opened 2017-08-20
Use proper icon for language switcher in navbar

Changaco:
https://github.com/liberapay/liberapay.com/issues/608#issuecomment-323548507 made me realize that there is a proper icon for language switchers, and that it's in Font Awesome.

### DIFF
--- a/templates/navbar-common.html
+++ b/templates/navbar-common.html
@@ -4,7 +4,7 @@
     <button class="dropdown-toggle navbar-btn" type="button"
             data-toggle="dropdown"
             title="{{ _('Switch to another language') }}">
-        <span class="glyphicon glyphicon-map-marker" aria-hidden="true"></span>
+        <span class="fa fa-language" aria-hidden="true"></span>
         <span class="text">{{ locale.language.upper() }}</span>
         <span class="caret"></span>
     </button>


Changaco:
Screenshots:

liberapay-2017-08-20-language-icon-md-screen

---

liberapay-2017-08-20-language-icon-xs-screen

Changaco:
It's a more complex icon so it doesn't exactly blend in with the others. Should we use it anyway or keep the old one?

vv221:
I’m not sure I would have guess it is used for language switching in the icon-only setup if I had not already seen it on the icon+text setup. I guess it is a bit too complex to be easily understandable in the size it is displayed.

jorgesumle:
The borders of the others icons are a bit thick, and this one is very thin. It should be proportional.

Changaco:
This got me wondering what icons Bootstrap 4 has, and the answer is: none. https://getbootstrap.com/docs/4.0/extend/icons/

Changaco:
https://joinmastodon.org/ simply uses the two symbols "文A" instead of the icon.

screenshot-2017-9-28 mastodon

jorgesumle:
`文A DE`, for example, will look confusing, because we have letters after the icon. An icon is better in our case I think.

PR #679, opened 2017-08-27
Improve payment card form

Changaco:
I've spent way too much time on this but I think the result is pretty good.

Screenshots:

spectacle ns9544

---

spectacle ns9608

---

spectacle wz9608

---

spectacle bc9608

---

spectacle fw9608

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -2,14 +2,16 @@ Liberapay.forms = {};
 
 Liberapay.forms.clearInvalid = function($form) {
     $form.find('.invalid').removeClass('invalid');
+    $form.find('.abnormal').removeClass('abnormal');
 };
 
 Liberapay.forms.focusInvalid = function($form) {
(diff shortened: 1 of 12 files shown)


Changaco:
I tested this in two WebKit browsers and was very confused because the "flex" wasn't working properly. Turns out `display: flex` doesn't work on `<fieldset>` elements in WebKit: <https://github.com/philipwalton/flexbugs#9-some-html-elements-cant-be-flex-containers>.

Changaco:
I've written a bit of documentation for `payment-cards.js` and pushed it to its own repository: https://github.com/liberapay/payment-cards.js

Changaco:
In this PR I accidentally dropped a JavaScript function that was still used. This bug prevented users who hadn't already added a bank account from withdrawing money. Fix: e403a94429df999884bd494bd607d7cdfbc2d3a2.

PR #686, opened 2017-09-05
Write 'Liberapay' string in upper case

jorgesumle:
To keep consistency.


Changaco:
Done in ca5cdd633caed5fed315d5063c8d563f1862445e.

Issue #678, opened 2017-08-27
Toggle navigation doesn't work when JavaScript is disabled 


jorgesumle:
This is because Bootstrap doesn't provide a fallback.

![captura de pantalla de 2017-08-27 13-02-05](https://user-images.githubusercontent.com/14212780/29749282-4afe3198-8b28-11e7-9575-92a37ef9d784.png)

jorgesumle:
Making it work without JavaScript is not easy, but I found a hacky solution. I explain it in the article [Barra de navegación de Bootstrap sin JavaScript](https://www.freakspot.net/barra-de-navegacion-de-bootstrap-sin-javascript/) (currently only in Spanish). You can see how it works here: https://freakspot.net/wp-content/uploads/2017/09/barra-navegacion-Bootstrap/con-y-sin-js.html

Basically, I hide the `<button>` when JavaScript is disabled and show a `<label>` which is associated with a hidden checkbox. When the checkbox is checked, the dropdown is shown.
```
#navbar-toggle-cbox:checked ~ .collapse {
          display: block;
}
```

Changaco:
Nice one. It should work in pretty much all browsers, except IE ≤ 8 (which we already don't support 'cause it's an obsolete piece of crap). http://caniuse.com/css-sel3

PR #706, opened 2017-09-20
Make toggle button from navigation bar work without JavaScript

jorgesumle:
Solves https://github.com/liberapay/liberapay.com/issues/678

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -11,6 +11,13 @@
     <link rel="stylesheet" href="{{ website.asset('bootstrap/css/bootstrap.css') }}">
     <link rel="stylesheet" href="{{ website.asset('base.css') }}" type="text/css">
     <link rel="stylesheet" href="{{ website.asset('fontawesome/4.5.0/css/font-awesome.min.css') }}" type="text/css">
+    <noscript>
+        <style>
+            label[for=navbar-toggle-cbox] { cursor: pointer; }
+            #navbar-toggle-cbox:checked ~ .collapse { display: block; }
+            #toggle-navbar-link { display: none; }
+        </style>
+    </noscript>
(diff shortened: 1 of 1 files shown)


jorgesumle:
> inline styles are a bad practice, they make pages heavier

Yeah, it would be better add the code to the file: `navbar.scss`

> I see no reason to keep using Bootstrap's JavaScript here, we can drop all the data- attributes and all the <noscript> tags

I'm  also OK not using Bootstrap JavaScript, but since we are already loading `bootstrap.js` and jQuery. If we don't use those JS files, we will lose the animation (we could create another one with only CSS though) and it will be impossible to change the `aria-expanded` to true. The aria-expanded changes its value when JavaScript enabled, but when JavaScript is disabled it can have a wrong value, as it cannot change. This is the only reason I used the `<noscript>` tag, and I wish I could have better a solution for that, but as far as I know it's impossible to change the value of an attribute dynamically without JavaScript. :(

Indeed, the data attributes are not necessary for the code inside the `<noscript>` tag.

> Code duplication is also a bad practice, there should only be one element that toggles navigation

One is a label which one appears when JS is disabled, the other is a link which is visible when JS is enabled. I don't see any duplication, they do different things and are never shown together.

> shouldn't role="button" be on the label instead of on the checkbox?

This one is very interesting. The checkbox is used as a button, but as the hidden elements are not read by screen readers, it would make more sense to add that attribute it in the label. Then we would have to add the `tabindex` attribute to make the label focusable.
---
There must be a better solution... The code is not easy to understand at first sight.

jorgesumle:
There is a much cleaner approach using the `:target` pseudoelement I learnt from @mosra and their Pelican theme [m.css](https://github.com/mosra/m.css), so I'm closing this unfinished and dirty pull request.

PR #708, opened 2017-09-22
Implement redirects for renamed profiles

### LINKED ISSUE #22: Handle username changes better
- [ ] throttle them
- [ ] implement profile redirection (see #91)

### LINKED ISSUE #91: Implement redirections based on a DB table
URLs don't stay the same forever, they can be changed by a user (e.g. #22) or by us (e.g. #90), and we don't want to clutter the source code with various ways of redirecting users to the right place, so:
- [ ] create a `redirections` table
- [ ] add a function to the website algorithm that queries the table when we're about to return a 404 to the client, and returns a 302 instead if there is a match

Changaco:
This branch closes #22 and closes #91. Instead of trying to redirect every 404 request in general (see the 2nd item of #91), which would cost an extra DB query for each 404, we only try to redirect unknown usernames in the `get_participant()` function.

### DIFF
--- a/liberapay/constants.py
+++ b/liberapay/constants.py
@@ -195,6 +195,7 @@ QUARANTINE = timedelta(weeks=4)
 RATE_LIMITS = {
     'add_email.source': (5, 60*60*24),  # 5 per day
     'add_email.target': (2, 60*60*24),  # 2 per day
+    'change_username': (7, 60*60*24*7),  # 7 per week
     'log-in.email': (10, 60*60*24),  # 10 per day
     'log-in.email.not-verified': (2, 60*60*24),  # 2 per day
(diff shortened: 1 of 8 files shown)


Changaco:
The last commit fixes an unrelated problem introduced in #703.

Changaco:
This branch doesn't add redirections for renames that were done in the past, here are some numbers on those:

```sql
select renames, count(*) as participants from (
    select (count(*) - 1) as renames from events where type = 'set_username' group by participant
) a where renames > 0 group by renames;
 renames | participants 
---------+--------------
       1 |           53
       2 |            8
       3 |            3
(3 rows)
```

(53 users have renamed their account once, 8 have done it twice, etc)

PR #716, opened 2017-09-26
Improve the speed of the test suite

Changaco:
With the simple change in this branch I'm seeing a gain of more than 35% on my machine (~59 seconds vs ~94 seconds before).


Changaco:
The speedup appears to be similar on Travis, between 30% and 50% faster.

PR #731, opened 2017-10-10
Redesign profile page

Bat41:
Fixes #608 

It's not exactly the same as the mockup I did, but it is still very similar. Here is how it looks like.

screen shot 2017-10-10 at 17 24 51-fullpage

I also changed the font to Open Sans. I don't know if it's fine, or if we should stick to Ubuntu.

PR #728, opened 2017-10-09
Improve profile boxes

Changaco:
This branch modifies the lists of users in the Explore pages, and revamps the elsewhere pages. It's the continuation of #726.

Issue #434, opened 2016-11-04
Actual takes are not cached, the estimations are wrong


Changaco:
Reticketing from #107. Most of the problems have been fixed but there is one left: the python code does a single loop that doesn't account for the impossibility of self-giving through a team, while the pgSQL code in payday uses a nested loop.

The solution is to cache actual takes in the DB instead of re-estimating them on each request.

Issue #21, opened 2015-08-30
Run `Payday.update_cached_amounts` regularly


Changaco:
To improve the accuracy of the cached giving/receiving numbers. We already have a cron mechanism, so adding another task should be pretty easy.

PR #739, opened 2017-10-13
Count team patrons

Changaco:
With #728 and #731 we are starting to show the number of patrons (donors) in addition to - or as a substitute for - the income. Unfortunately counting patrons is a little complicated in Liberapay because of teams. This branch creates a column named `nteampatrons` to complement the existing `npatrons` column, and uses it in the new profile boxes and headers.

### DIFF
--- a/liberapay/billing/payday.py
+++ b/liberapay/billing/payday.py
@@ -498,8 +498,9 @@ class Payday(object):
             log(msg % (t.id, t.amount, t.context, t.team, t.tipper_wallet_id, t.tippee_wallet_id))
             transfer(db, **t.__dict__)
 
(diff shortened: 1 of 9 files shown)


Changaco:
After another review I realized that this commit alone isn't enough to fix the issue of the displayed patrons count being zero alongside a non-zero income, so I'm going to push more stuff into this branch. We can't determine the exact patrons count without running a full virtual payday, but we do know that if we have $x patrons from one team then the total `nteampatrons` for all teams that the user belongs to is at least $x, and that will fix our "zero patrons" problem.

Changaco:
Ready for review. The new commits also take care of #434 and #21.

PR #754, opened 2017-10-20
Update tip-distribution.html

trebmuh:
Add a space for a better readability.

Trying to go from here ( https://en.liberapay.com/$USER/receiving ):
capture 1

To something like that:
capture 2

@Changaco : I'm not really sure of what I'm doing (speaking about code) so, please, double-check it!

### DIFF
--- a/templates/tip-distribution.html
+++ b/templates/tip-distribution.html
@@ -2,7 +2,7 @@
         % for amount, ncontributors, summed, pcontributors, psummed in tip_distribution
         <tr>
             <td class="currency-symbol">€</td>
-            <td class="amount">{{ amount }}</td>
+            <td class="amount">{{ amount }} </td>
             <td class="count">
                 <span class="bar" style="width:{{ pcontributors * 240 }}pt"></span>
                 {{ "%5.1f" % (pcontributors * 100) }}%


Changaco:
The proper way to fix this is to bring back the style for `.bar`.

Changaco:
Here is the original style which was deleted in 40eb9737750e02a9909cd68639a2af2c68c447c9:

```scss
.tip-distribution {
    font: normal 12px $Mono;
    .dollar-sign {
        padding: 0 2pt 0 24pt;
        text-align: right;
    }
    .amount {
        padding: 0 6pt 0 0;
        text-align: right;
    }
    .amount-change {
        padding: 6pt 0 6pt 24pt;
        text-align: left;
    }
    .count {
        text-align: left;
        white-space: nowrap;
    }
    .count span.number {
        font-size: 8pt;
    }
    .count span.bar {
        background: $light-brown;
        display: inline-block;
        margin-right: 3pt;
        height: 9pt;
    }
    .count span.bar.green {
        background: $green;
    }
}
```

trebmuh:
I'm not that well with CSS unfortunately. I'm leaving that to other then.

@Changaco would you like me to close this one and open an issue about it instead?

Changaco:
@trebmuh Yes, I guess an issue would be better.

trebmuh:
Done. Closing here now.

Issue #758, opened 2017-10-22
Wording is not clear enough when adding money


Changaco:
https://twitter.com/josh_cheek/status/921716393766211584

ziyadedher:
I am interested in this, but I have been looking around trying to find where the error messages are stored and they seem to be in a database somewhere which I cannot find.  Could you lead me in the right direction?

Changaco:
Actually this issue isn't about the error message, it's about the wording of the page before that, which apparently doesn't make it clear that the payment is attempted immediately and in full, as opposed to a weekly charge.

PR #763, opened 2017-10-25
Clarify when transfers happen

mattbk:
Closes https://github.com/liberapay/liberapay.com/issues/758.


mattbk:
Something like this? Or should there be more information?

Changaco:
Only card payments are immediate. Direct debits happen 4 days later ([doc](https://docs.mangopay.com/endpoints/v2.01/payins#e281_create-a-direct-debit-web-payin)). As for bank wires we have no control over when the payment is actually initiated, it's up to the user and their bank.

Regarding the sentence you added, I'm worried it could be interpreted as "the charge will happen as soon as you select or input an amount", which isn't how this form works.

mattbk:
Ooooh, good points.

Changaco:
Thanks for trying @mattbk. Hopefully someday we'll find the right wording.

Issue #182, opened 2016-02-23
Implement multiple currencies


Changaco:
I just realized that we didn't actually have an open issue for this, so here it is.

MangoPay [currently supports](https://www.mangopay.com/features/) the following currencies in addition to euros: GBP, CAD, SEK, NOK, DKK, CHF, ZAR, PLN, USD, AUD.

[Fees](https://www.mangopay.com/pricing/pricing-worldwide/) become a lot more complicated when we introduce other currencies into the equation.

Changaco:
I think this is still not a priority (previous statement: https://github.com/liberapay/liberapay.com/issues/44#issuecomment-159319824).

Changaco:
I asked MangoPay if they were planning to add support for currency conversion to their API, they said no.

Not having currency conversion means that it would be more likely that a donor could find themselves in the annoying situation of having money in their Liberapay account but still needing to add more because the person they want to give to only accepts other currencies.

MartinDelille:
Would it be possible to rely on another service for conversion?

Changaco:
> Would it be possible to rely on another service for conversion?

Assuming that such a service exists, it would be complicated and probably not instantaneous. There would also be the problem of bypassing MangoPay's fees.

We could allow users to trade currencies between themselves, but that wouldn't guarantee always being able to convert your money when you want to.

aggsol:
Does this include Bitcoin (XBT)?

Changaco:
> Does this include Bitcoin (XBT)?

Not directly, but implementing native support of Bitcoins requires multiple currencies accounting. That's why #364 is tagged as Blocked on this issue.

aggsol:
If the incoming money would always exchanged to Euro then no multiple currency accounting would be necessary. Like the the provider would accept Bitcoin, USD etc. on Liberapays behalf, exchange it to Euros.

Changaco:
Automatic currency conversion is built into credit card systems, Liberapay donors whose bank accounts aren't denominated in euros can pay by card. This issue is about implementing multiple currencies in Liberapay itself.

aggsol:
Just to be clear: Running everything in EUR and just displaying a converted amount in an another currency is not an option?

Changaco:
No, we want real support for multiple currencies, it's the big missing piece of our internationalization efforts.

Changaco:
The priority is US dollars, and we can start by only supporting one currency per account if that simplifies things.

Changaco:
[...]

PR #768, opened 2017-10-29
Currencies - part 2

Changaco:
Second step towards #182.


Changaco:
I didn't get a response from Mangopay regarding fees on USD transactions, so I went with an exchange rate of 1.2, which gives the following fees (incl. VAT):

- 2.925% + $0.35 for card payments
- $3.50 for withdrawals

I've set the minimum charge to $20 for cards, which results in a 4.58% fee.

PR #778, opened 2017-11-02
updated link to Snowdrift.coop market-research

wolftune:
the old one redirects still, but this skips the redirect


Changaco:
Thanks @wolftune. :-) I see you've already removed Gratipay from the list!

Issue #800, opened 2017-11-12
The anonymity of donations is not explained enough


Changaco:
It's only mentioned in <https://liberapay.com/about/>. We should at least add a message on users' Receiving page (<https://liberapay.com/about/me/receiving/>).

Changaco:
I guess this can be considered solved now that non-anonymous donations have been implemented (#2088).

Issue #271, opened 2016-04-17
Anonymity prevents thanking sponsors


Changaco:
Even if we explicitly allowed users to thank their sponsors (https://github.com/liberapay/liberapay.org/issues/12) they couldn't do it because donations are anonymous. We need to implement a way for donors to reveal their identity if they and their donees want to.

Changaco:
+2 from https://www.peppercarrot.com/en/article418/peppercarrot-is-now-on-liberapay#c0418-1468

bitJericho:
I'd love this feature. With Patreon in the news, now would be the perfect time to implement this feature which is a big pain point for users switching from there.

mattbk:
-1 from me. I think it's a legal bag of worms.

gregzaal:
> legal bag of worms.

Forgive my ignorance, but why? Would an opt-in to be recognised (but default to anonymous) solve this?

mattbk:
Sorry, I posted in the wrong place. Carry on.

mattbk:
Sorry, I posted in the wrong place.

Although I'm personally in favor of anonymous donations because then donors can't demand things in return for their donations ("I give you $10 a month, make me a sandwhich."), is there any reason not to do this?

trebmuh:
The "pure donation" vision is the main "why" I came first to liberapay.

I've nothing against introducing a "patreon/tipeee/..." business model as long as the "pure donation" thing is kept possible and clear enough (I don't want to have nationstate-taxes coming to me thinking I'm making money of a project).

@Changaco : is LP still on this line?

Changaco:
Yes, pure donations will always be possible on Liberapay.

trebmuh:
Thanks for the clarification @Changaco 👍

davidak:
Could there just be a list of supporters on a users public profile with the last point "and X anonymous supporters"? Maybe hide it by default for simplicity and make it visible by clicking on some button "show supporters".

Also, a user might want to show which projects they support publicly. Like on crowdfunding pages.

rugk:
…or implement an anonymous way to thank sponsors. So sponsored people would have a button "Thank my sponsors", where they can write a message to all sponsors – or maybe only to the new once since their last message or so.

ashledombos:
[...]

PR #805, opened 2017-11-20
 Add visibility of giving and receiving

oliverpool:
Related to https://github.com/liberapay/liberapay.com/issues/800.

This adds a link near the text `You give/receive {amount} per week` mentioning the visibility (`public` or `secret`), with a link to the settings page (to eventually change it).

---
2017-11-20-153509_2560x1080_scrot

---
2017-11-20-153605_2560x1080_scrot

---

This is my first actual contribution to liberapay, so please feel free to point out my mistakes or where I could improve the code (especially regarding how the translation should be handled)

### DIFF
--- a/www/%username/giving/index.html.spt
+++ b/www/%username/giving/index.html.spt
@@ -69,7 +69,14 @@ weekly = total - participant.receiving
         % endif
     </p>
 % else
-    <p>{{ _("You give {0} per week.", participant.giving) }}</p>
+    <p>
+        {{ _("You give {amount} per week ({link_open}{visibility}{link_close}).",
+            amount=participant.giving,
+            link_open='<a href="%s">'|safe % participant.path('settings'),
+            visibility=(_("hidden") if participant['hide_giving'] else _("public")),
+            link_close='</a>'|safe,
+        ) }}
+    </p>
 % endif
 
 % if not freeload and not Liberapay_tip.amount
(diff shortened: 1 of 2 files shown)


Changaco:
I'm not convinced by the proposed UI change, I think it's distracting and potentially confusing. What are you trying to accomplish by directing the user's attention to the privacy settings?

oliverpool:
I think I misread the issue #800 and thought that you were talking about the anonymity of the amount given/received (when you actually meant that your benefactors are anonymous).

Anyway, I think it is important to mention to the user if the amount they receive/give is secret or public (the link is there just to show the user that they can control this visibility parameter)

whit537:
Would this look better?

> You [publicly]() receive €0.00 per week.
> You [secretly]() give €0.00 per week.

I agree the UI could be improved but tbh that's somewhat true across the board. Not sure this change actively degrades the existing experience.

Changaco:
I remain unconvinced and I think a single word isn't clear enough, especially in the long run with the possibility of non-anonymous donations (#271). Something like "(This information is public.)" might work.

oliverpool:
> especially in the long run with the possibility of non-anonymous donations

Then maybe a specific "is public"-icon would be nice to quickly know if an information is public (I would personally assume that everything that isn't explicitly marked as "public" is "private/secret").

What do you think of the [bullhorn](http://fontawesome.io/icon/bullhorn/)? :smile:

MartinDelille:
@oliverpool The bullhorn evocate a notification. For mem the icon indicating that an information is public is the [earth](https://fontawesome.com/icons/globe-africa?style=solid) (maybe because I use facebook a lot).

PR #807, opened 2017-11-21
Autodetect pledge platform

### LINKED ISSUE #806: [feature] autodetect social network of an account
This is a minor feature that could be nice and easy to implement:

When one start a pledge 

image

Autodetect (by a simple test on regular expressions `/http*://twitter.com/`, `http*://github.com/`) the social network on which a user is registered,

and remove the beginning of the url and change the value of the select (Github to twitter in the example)

Bat41:
Fixes #806 

And BTW, maybe we could use Babel to transpile our JS while keeping compatibility with old browsers (I miss arrow functions, `const` and `for (... in ...)` so much).

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -72,3 +72,22 @@ Liberapay.forms.success = function($form, $inputs, button) { return function(dat
         window.location.href = window.location.href;
     }
 }};
+
+Liberapay.forms.platforms = {
+    'twitter': /^(https?:\/\/)?(www\.)?twitter\.com\/(\w+)/,
+    'github': /^(https?:\/\/)?(www\.)?github\.com\/(\w+)/,
+    'bitbucket': /^(https?:\/\/)?(www\.)?bitbucket\.org\/(\w+)/
+};
+
+Liberapay.forms.detectPlatform = function () {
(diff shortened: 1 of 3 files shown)


whit537:
I haven't tested this, but it looks reasonable at a skim and I don't smell any security issues and it's small. I suggest merging to keep up velocity.

whit537:
Also, transpilers scare me. Too much complexity! 🙈

Changaco:
I didn't merge this immediately because it adds JavaScript debt.

I think the proper way to fix #806 is to come up with a better UI for pledgee lookup, not add a hidden feature which allows you to paste a URL in a username input.

Changaco:
The future of pledgee lookup is to do a fuzzy search when the input is a name, instead of redirecting to a page which can be a 404.

For this PR I suggest increasing the width of the input and changing its placeholder to "username or URL", as well as moving the URL detection server-side to avoid JavaScript debt.

Changaco:
Or maybe a separate input for URLs.

Bat41:
If you prefer to process the input on the server, I'll let you re-implement this, as I feel really uncomfortable with Pando.

PR #685, opened 2017-09-04
Implement direct debits

Changaco:
This branch implements a new payment method: SEPA direct debits (closes #57). It also improves a few related things, and fixes at least one bug.

PR #809, opened 2017-11-23
Fix payment routes

Changaco:
The first commit should have been in #645, and the second in #685.


Changaco:
I found another bug while testing in sandbox. It's not related to payment routes but I've put the fix in this branch anyway.

PR #814, opened 2017-11-29
Support unicode characters in emails

### LINKED ISSUE #766: Email addresses with non-ascii characters are rejected
https://twitter.com/transposees/status/923842688948101121

BaptisteGelez:
I saw that the previous regular expression was the same Mangopay's one, I don't know if the new one is still valid (I couldn't find the regular expression they are using to check emails).

Fixes #766

### DIFF
--- a/liberapay/models/participant.py
+++ b/liberapay/models/participant.py
@@ -704,7 +704,7 @@ class Participant(Model, MixinTeam):
             Returns the number of emails sent.
         """
 
-        # normalize the address: strip it, lowercase the domain name, and
+        # normalize the address: strip it, then lowercase and encode the domain name
         email = email.strip()
         i = email.rfind('@') + 1
         email = email[:i] + email[i:].lower().encode('idna').decode()
--- a/liberapay/models/participant.py
+++ b/liberapay/models/participant.py
@@ -704,10 +704,10 @@ class Participant(Model, MixinTeam):
             Returns the number of emails sent.
         """
 
-        # normalize the address: strip it, and lowercase the domain name
+        # normalize the address: strip it, lowercase the domain name, and
         email = email.strip()
-        i = email.rfind('@')
-        email = email[:i] + email[i:].lower()
+        i = email.rfind('@') + 1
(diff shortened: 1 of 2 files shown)


payday-1.txt
```diff
@@ -0,0 +1 @@
+fake log file
```
Bat41: Should I delete this file?

Changaco: Yes, and the other one too.

Bat41:
Is it possible to use Punycode for the whole address, or is it limited to domain names?

Changaco:
Punycode is only for domain names. Hint: the python codec name is `idna`.

liberapay/models/participant.py
```diff
@@ -645,10 +645,10 @@ def add_email(self, email, cursor=None):
             Returns the number of emails sent.
         """
 
-        # normalize the address: strip it, and lowercase the domain name
+        # normalize the address: strip it, lowercase the domain name, and
```
whit537: Sup with the `, and` move?

tests/py/test_email.py
```diff
@@ -46,6 +48,11 @@ def test_participant_can_add_email(self):
         response = self.hit_email_spt('add-email', 'alice@example.com')
         assert response.text == '{}'
 
+    def test_participant_can_add_email_with_special_chars(self):
+        punycode_email = 'alice@' + u'accentué.com'.encode('idna').decode()
+        self.hit_email_spt('add-email', u'alice@accentué.com')
+        assert self.alice.get_any_email() == punycode_email
```
whit537: Can we see special chars in the username part as well? What happens if we pass in `alïçê@accentué.com`?

Bat41: It should normally throw an error, as this adress doesn't respect the regular expression we are using for validation. If Mangopay doesn't fully support accented emails, I think there is no way to do it on our side.

whit537: @Bat41 Can we encode that knowledge in a test? :)

PR #828, opened 2017-12-05
Refactor Shell Scripts based on ShellCheck's suggestion

PeterDaveHello:
This will help improve the robustness of the scripts, some of the deprecated and non-safe statements need to be refactored.


Changaco:
Thanks but I'm not going to merge this as-is. The changes seem mostly unnecessary, and some of them are in obsolete files that I should have deleted months ago.

PeterDaveHello:
I can revert the changes on files need to be deleted. I don't think they are all that unnecessary, for example, one of their header used `#!/bin/sh` which may not support many of statements in the script, and `-o`/`-a` in bash condition were deprecated for a while, it's no hurt to have these changes, they only make those scripts less potential of issues.

whit537:
Merging the changes to the obsolete files makes no difference since the files are obsolete anyway.

Merging the unnecessary changes probably(?) won't hurt.

I agree that @PeterDaveHello's PRs such as this one tend to be low-value, but ¯\\\_(ツ)\_/¯. Either merge or close to keep the PR queue moving.

Changaco:
Rebased on master and added a commit to delete the obsolete files.

Changaco:
And another to revert changes that weren't merely unnecessary but invalid.

release.sh
```diff
@@ -1,13 +1,13 @@
-#!/bin/sh
+#!/bin/bash
 
 # Fail on errors and undefined variables
 set -eu
 
 # Be somewhere predictable
-cd "`dirname $0`"
+cd "$(dirname "$0")"
```
whit537: I don't know how the double quotes behave here.

whit537:
Assuming those double double quotes lines work as intended, lgtm.

release.sh
Changaco: This is quite common in shell scripts, I think it works as intended.

PR #842, opened 2017-12-09
Change Twitch OAuth scope to user_read

### LINKED ISSUE #836: Overly broad permissions requested for Twitch accounts
Copy-paste from an email we received yesterday:

> Fix your twitch linking to NOT ask for access to stream key. That is not something you need one way or another.

My response:

> We try to ask for the minimum amount of permissions when linking accounts from other platforms, so if we're getting more from Twitch than necessary then it's because it's probably impossible to ask for less (or at least it wasn't possible when we added support for this platform).

riking:
This still provides the email, without a scary warning or exposing the stream key.

Docs: https://dev.twitch.tv/docs/authentication#scopes

Fixes #836 

This is not tested since I don't feel like setting up a development environment complete with Twitch API key, but the change should be obvious enough.

### DIFF
--- a/liberapay/elsewhere/twitch.py
+++ b/liberapay/elsewhere/twitch.py
@@ -39,7 +39,7 @@ class Twitch(PlatformOAuth2):
     # Auth attributes
     auth_url = 'https://api.twitch.tv/kraken/oauth2/authorize'
     access_token_url = 'https://api.twitch.tv/kraken/oauth2/token'
-    oauth_default_scope = ['channel_read']
+    oauth_default_scope = ['user_read']
     session_class = TwitchOAuthSession
 
     # API attributes


Changaco:
Unfortunately this doesn't work, because we use the `/kraken/channel` endpoint to get the data we need and it requires the `channel_read` permission.

whit537:
So should this be closed/changed in favor of a doc change? Explain to the user why the permission is being requested?

Changaco:
Yes, if we can't properly fix the permissions then we should add an explanation message. We should also drop the stream key from the data we receive so that it doesn't end up in our database.

whit537:
Okay, those sound like separate PRs. Suggest closing this one to keep the queue clean.

PR #848, opened 2017-12-09
Change amount inputs to type "number"

### LINKED ISSUE #786: Donation should be number pad on mobile and text area should be number
At least on my phone (Android Marshmallow), this doesn't happen and I get an automatic space after entering ".".

Same as https://github.com/gratipay/gratipay.com/issues/4593.

pottedmeat:
Input fields that accept a numeric amount should use `type="number"` to enable devices with software keyboards to automatically display a number pad. This also prevents keyboard auto-insertion such as the addition of a space character after pressing `.` when entering decimals.

Using this attribute also introduces a "step" arrow within the input field on devices with pointers. By default, this increases the numeric value +1/-1 which is worth discussing if anyone feels this behavior is unwelcome.

Resolves #786.

### DIFF
--- a/templates/your-tip.html
+++ b/templates/your-tip.html
@@ -29,7 +29,7 @@
     <div class="input-group">
         % set currency = tip.amount.currency
         <div class="input-group-addon">{{ locale.currency_symbols.get(currency, currency) }}</div>
-        <input type="text" name="amount" id="amount" placeholder="{{ _('Amount') }}"
+        <input type="number" name="amount" id="amount" placeholder="{{ _('Amount') }}"
                class="amount form-control {{ 'input-sm' if small else '' }}"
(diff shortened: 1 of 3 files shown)


Changaco:
I'm pretty sure we don't want to use the `number` type, it's weird looking and can potentially conflict with i18n if the browser is stupid or badly configured.

I vaguely remember reading something about using `tel` instead to get the number pad, while disabling input validation since we don't actually want a phone number.

pottedmeat:
That makes some sense. Let me see if I can find some good documentation for it.

pottedmeat:
This is a great analysis: https://www.ctrl.blog/entry/html5-input-number-localization

whit537:
@pottedmeat So what's the upshot? New commit here?

Changaco:
Closing in favor of #1007.

PR #857, opened 2017-12-12
Clarify that Liberapay is not self-hostable

Changaco:
Some people assume that because Liberapay is open source they can run their own instance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@
 
 [Liberapay](http://liberapay.com) is a recurrent donations platform. We help you fund the creators and projects you appreciate.
 
+Note: This webapp is not self-hostable.
+
 ## Table of Contents
 
 - [Contact](#contact)
@@ -164,6 +166,8 @@ We use [pip's Hash-Checking Mode](https://pip.pypa.io/en/stable/reference/pip_in
 If for some reason you need to rehash all requirements, run `make rehash-requirements`.
 
 ### Deploying the app
+
+Note: Liberapay cannot be self-hosted, this section is only meant to document how we deploy new versions.
 
 Liberapay is currently hosted on [AWS](https://aws.amazon.com/) (Ireland).
 


revi:
What about saying "This app is not designed to be self-hosted", rather than it's not self-hostable? With ~~more than enough~~ diligence they can self-host. Well... I think that's de-facto forking, though.

Changaco:
I think "not self-hostable" is appropriate since anything hosted by someone else wouldn't actually be Liberapay.

saschafoerster:
I would be interested in an explication, why can't you install Liberapay on your own server!

Changaco:
Liberapay is a centralized platform that acts as a trusted third party, there can only be one of it.

saschafoerster:
We are building local platforms around the city of Bonn. So trust means for me, I can meet the administrator in person, look into his/hers eyes. So I would love to build a local crowdfunding-platform and I find liberapay very appealing to fund projects like https://Bonn.social (based on Mastodon). So maybe you could think of not just making liberapay opensource but really set it free with several instances build be different administrators who are trusted by their communities. :) So, this is some more background to my question, maybe you can help me understand your position better as well.

mattbk:
The [license is](https://github.com/liberapay/liberapay.com#license) CC0...you can do what you want with it.

bearbin:
I this is really a misunderstanding between the software and the platform. You can run the software Liberapay on your own infrastructure, connecting to your own payment providers and with its own name (not "Liberapay"), and it will work fine. It is free and open source.

However, you cannot self-host the service "Liberapay" - that is run by the Liberapay organisation, which is incorporated in France and is a centralised SaaS using private relationships with payment processors. The trademark "Liberapay" is associated only with the organisation, and there is no way to interact with other liberapay instances at all.

saschafoerster:
So, for instance, if I think "Bundescrowd.de" as a name would be a great platform only for local projects, I could use and install the software labelled "Liberapay without big problems. I will try it in my next free playtime! Of course, some help with installing this software would be great, even though "self-hosting" is not wished for, as I feel.

Changaco:
I will not make any effort to facilitate the creation of other "instances" of Liberapay unless someone can convince me that such offshoots are a good and viable idea. Liberapay is not Mastodon, it's a payment service that handles real money and personal data, not a simple micro-blogging service, it can't be federated like Mastodon is.

Issue #788, opened 2017-11-06
Feed or drill-down on Recent Activity


mattbk:
![image](https://user-images.githubusercontent.com/3729038/32454831-ffe85c06-c2e5-11e7-9a8f-e7f62f32645b.png)

The Recent Activity box on the front page makes me ask all sorts of questions: Who are these people, and should I donate to them? Can I see how the funding changes mid-week? Etc.

Would be nice to at least have a feed of new accounts and pledges in addition to the leaderboards under /explore.

Changaco:
We used to list new accounts on the homepage, but without upfront moderation we quickly ended up "advertising" controversial accounts. Also without filtering by profile completeness the list isn't very useful to anyone but those who want to keep an eye on who is joining Liberapay, i.e. it's only interesting to us.

mattbk:
Aha, similar to https://github.com/liberapay/salon/issues/45 ?

It's an interesting problem. I'll keep thinking about how to avoid the moderation aspect of it. Obviously we couldn't show new donations and keep anonymity.

mattbk:
How about a breakdown for both month and week?

In the last week:
- n new user accounts
- n new donations
- n new pledges
- $ transferred

In the last month:
- n new user accounts
- n new donations
- n new pledges
- $ transferred

mattbk:
NB https://github.com/liberapay/liberapay.com/blob/87dacd58e9e29a9b5920cb1f2f211e105dd0ae56/www/index.html.spt#L21

mattbk:
![image](https://user-images.githubusercontent.com/3729038/33957670-e5ec7e56-e007-11e7-8304-506d2e79b16e.png)

Changaco:
Alternative idea regarding the "granularity" aspect of the issue: recent activity graphs showing the daily numbers for the past month, either in a new `/about/recent-activity` page or as a new section in `/about/stats`.

mattbk:
That's what I was thinking about yesterday afternoon, and what I would actually want. I just have no time to mess with it anymore.

PR #868, opened 2017-12-13
Revamp recent activity.

mattbk:
Part of https://github.com/liberapay/liberapay.com/issues/788.

Try number two...

Not sure if there is a way to just run one query, feedback appreciated.

### DIFF
--- a/www/index.html.spt
+++ b/www/index.html.spt
@@ -47,6 +47,35 @@ recent = query_cache.one("""
          , (SELECT nactive FROM _payday) AS newest_payday_nactive
 """, (currency, currency))
 
+recentweek = query_cache.one("""
+    WITH _users AS (
+             SELECT join_time
+               FROM participants
+              WHERE join_time > (now() - INTERVAL '7 days')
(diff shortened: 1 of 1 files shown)


mattbk:
image

- [ ] pull out most recent users into a separate line? Don't duplicate.
- [ ] should be monthly funding for month stats, not weekly (query looks right, text is wrong)
    - [ ] also for pledges
- [ ] flip week and month position (week first)
- [ ] rename monthly query

Changaco:
I think showing both past-week and past-month numbers at the same time is too much. A bit of JavaScript to allow the user to switch between the two time frames would be nicer. But really I'm not sure this kind of thing belongs on the front page, I don't see it being used by more than a small minority of visitors. I also don't think that this actually fixes #788 because your original message was mostly about being able to explore the recent activity.

mattbk:
Can I build it into an admin page?

Changaco:
Sure, feel free to build lots of admin tools, we need more of them.

PR #895, opened 2017-12-25
Update security page.

EdOverflow:
Add a clear description to our security policy.


jorgesumle:
Only people who speak English will understand it if we don't make translatable.

EdOverflow:
@jorgesumle, that is already required without this PR: https://fr.liberapay.com/about/security.

Changaco:
!m @EdOverflow

PR #896, opened 2017-12-25
Fix typo

jorgesumle:
Markdown is a proper noun


Changaco:
This change needs to be applied to all occurrences of "markdown" including in `i18n/`, otherwise all the translations will have to be redone even though the meaning of the message remains the same.

jorgesumle:
Done! I'm not sure about Korean, so I didn't change it. Maybe you have to remove the  이 from "마크다운이", but I don't really know.

Issue #688, opened 2017-09-07
HTTP API


Changaco:
We need a proper HTTP API so that external apps can interface with Liberapay, and for any deep integration we need to become an OAuth provider so that apps can access and/or modify a user's donations.

aggsol:
Why would any external app interface with Liberapay? What usecase do you see for creator and donors?

Changaco:
Some preliminary thoughts:

- Should the API be part of the main app? Or should it be a separate app?
- `https://api.liberapay.com/` or `https://liberapay.com/api`? The subdomain is probably better in the long run because it's easy to route it separately from the main app.
- Should we use [GraphQL](http://graphql.org/)? Here's an interesting comparison: [GraphQL vs. REST](https://dev-blog.apollodata.com/graphql-vs-rest-5d425123e34b).

aggsol:
I vote for: 
* a separated app makes it easier to move around, isolate, throttle etc
* `https://api.liberapay.com`
* REST is well know, easy to test (`curl` and `jq` ir more than enough) and liberapay is not complex enough IMO to benefit from GraphQL

Changaco:
https://github.com/postgraphql/postgraphql
https://github.com/begriffs/postgrest

Bat41:
I tried GraphQL once, and the first problem I had is there is no official solution for authentication. You can hack something, but it seems difficult to limit authentication to some resources. And as @aggsol said, GraphQL would probably be too much for Liberapay anyway.

timpritlove:
I want to voice my support for an API and I want to give an example what kind of software could integrate such an API: online banking software (e.g. MoneyMoney which provides an internal API itself <https://moneymoney-app.com/api/webbanking/>).

The goal would be to integrate Liberapay into banking apps as if they were a real bank. To the user, there is not a big difference:

- there is authenticated access to an account
- the account has a balance
- there are transactions going on (payouts and withdrawals)

So if LP could be treated like a bank account, it would automatically integrate into online banking software that is itself has an open API. An even more useful step would be to actually provide FINTS [1] compatibility which would make LP compatible with all online banking software in Germany.

However, a simple HTTP-based API that provides balance and uniquely identified transactions would be a good start.

[1] http://www.hbci-zka.de/english/

SirCmpwn:
[...]

PR #902, opened 2017-12-29
Minor API changes

Changaco:
This branch improves the `/%username/public.json` endpoint.

### DIFF
--- a/tests/py/test_public_json.py
+++ b/tests/py/test_public_json.py
@@ -70,7 +70,6 @@ class Tests(Harness):
     "avatar": null,
     "elsewhere": {
         "github": {
-            "id": %(elsewhere_id)s,
             "user_id": "%(user_id)s",
             "user_name": "bob"
         }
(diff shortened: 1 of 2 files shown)


whit537:
What's the reason for this change? I don't see a ticket linked.

I also don't find API doc on https://github.com/liberapay/liberapay.com/ or https://liberapay.com/ to know what's documented and not.

tests/py/test_public_json.py
```diff
@@ -70,7 +70,6 @@ def test_jsonp_works(self):
     "avatar": null,
     "elsewhere": {
         "github": {
-            "id": %(elsewhere_id)s,
```
whit537: Test the domain addition as well?

Changaco:
The reason for this change is that I was looking at `public.json` and it bothered me that the elsewhere data was incomplete, so I fixed it because it was simple to do. Then I realized the `id` keys are useless and confusing so I dropped them.

`public.json` is not documented. The ticket for a proper HTTP API with documentation is #688.

whit537:
> `public.json` is not documented.

In that case I don't see anything to worry about here. 👍

PR #917, opened 2018-01-09
Warn the user when the translation is incomplete

### LINKED ISSUE #712: Incomplete translation warning
It's not good that we display partially translated pages to users, but what's worse is that we don't display a warning about the translation being incomplete.

Changaco:
Closes #712.


Changaco:
Screenshots:

warning on small screen

---

warning on larger screen

Issue #756, opened 2017-10-21
Receiving page : improving readability


trebmuh:
On https://en.liberapay.com/$USER/receiving , we have now:
![31844638-493a8fb2-b5fa-11e7-86b9-d82aa09628e4](https://user-images.githubusercontent.com/8705846/31853359-5b34968a-b687-11e7-82c9-3e644c61b234.png)

It would be great to move to something like that:
![31844640-51e2c378-b5fa-11e7-9528-1e9a16018a46](https://user-images.githubusercontent.com/8705846/31853362-67716022-b687-11e7-9d0b-4a24fe5848de.png)

@Changaco made a suggestion here: https://github.com/liberapay/liberapay.com/pull/754#issuecomment-338376573 . Unfortunately, I'm not the one who could act here.

Issue #879, opened 2017-12-16
CSP could be more strict


Changaco:
Reticketing from <https://github.com/liberapay/liberapay.com/pull/498#pullrequestreview-83973507>.

- [ ] remove all inline scripts and drop `'unsafe-inline'` from `script-src`
- [ ] remove all inline styles and drop `'unsafe-inline'` from `style-src`
- [ ] restrict `connect-src` to `'self' %(main_domain)s *.mangopay.com *.payline.com`

Changaco:
Related story: [I’m harvesting credit card numbers and passwords from your site. Here’s how.](https://hackernoon.com/im-harvesting-credit-card-numbers-and-passwords-from-your-site-here-s-how-9a8cb347c5b5)

Changaco:
Current CSP of GitHub (homepage):

```
default-src 'none';
base-uri 'self';
child-src render.githubusercontent.com;
connect-src 'self' uploads.github.com status.github.com collector.githubapp.com api.github.com www.google-analytics.com github-cloud.s3.amazonaws.com github-production-repository-file-5c1aeb.s3.amazonaws.com github-production-upload-manifest-file-7fdce7.s3.amazonaws.com github-production-user-asset-6210df.s3.amazonaws.com wss://live.github.com; font-src assets-cdn.github.com;
form-action 'self' github.com gist.github.com;
frame-ancestors 'none';
img-src 'self' data: assets-cdn.github.com identicons.github.com collector.githubapp.com github-cloud.s3.amazonaws.com *.githubusercontent.com;
media-src 'none';
script-src assets-cdn.github.com;
style-src 'unsafe-inline' assets-cdn.github.com
```

Issue #504, opened 2017-01-10
Prevent 401 phishing attacks


EdOverflow:
The following markdown image should prompt you with an authentication dialog:

~~~
![](https://httpbin.org/basic-auth/user/passwd)
~~~

This can be used for 401 phishing attacks, where the credentials are sent to an attackers' server.

EdOverflow:
Maybe we should also prevent use of SVG images.

~~~
![](http://brutelogic.com.br/poc.svg) <-- Open this image in a new tab.
~~~

Changaco:
Nice find. :-) What solution would you recommend? The only effective one I can think of is to proxy all images. That's pretty much #202.

Changaco:
Note that Firefox does show a warning:

> https://httpbin.org is requesting your username and password. WARNING: Your password will not be sent to the website you are currently visiting!

Screenshot:

![spectacle h13774](https://cloud.githubusercontent.com/assets/1581590/21811897/893c17e4-d751-11e6-9ea3-fbe81d0a1c9c.png)

Changaco:
Can't we tell the browser not to show the prompt through CSP or a similar mechanism?

Why do browsers even show a prompt in this case? What legitimate use-case is there for this?

EdOverflow:
> Why do browsers even show a prompt in this case? What legitimate use-case is there for this?

This is the authentication mechanism for the HTTP Basic Authentication Protocol: http://blog.stevensanderson.com/2008/08/25/using-the-browsers-native-login-prompt/

Changaco:
I know what it is, but I see no reason to allow it for images. ;-)

EdOverflow:
Oh, sorry. There isn't really a legitimate use of this mechanism in images. For some reason developers are not interested in fixing this issue.

The only real solution is to proxy all images as suggested in #202.

Changaco:
> For some reason developers are not interested in fixing this issue.

Do you have links to the relevant tickets in the bug trackers of browsers?

EdOverflow:
> Do you have links to the relevant tickets in the bug trackers of browsers?

https://bugs.chromium.org/p/chromium/issues/detail?id=21628

> Can't we tell the browser not to show the prompt through CSP or a similar mechanism?

I will look into this. There is a discussion concerning a CSP implementation here: https://www.w3.org/2011/webappsec/track/issues/68

Changaco:
[...]

PR #924, opened 2018-01-12
CSP++

Changaco:
This branch tightens our Content Security Policy (#879). It also fixes a number of issues I came across, including #756.


Changaco:
CSP changes summary:

- dropped `'unsafe-inline'` from `script-src` and `style-src`
- added `form-action 'self'`
- restricted `connect-src` to `'self' %(main_domain)s *.liberapay.org *.mangopay.com *.payline.com`
- dropped `reflected-xss`

Remaining TODO in the future:

- restrict `img-src` (requires #504)

Changaco:
Screenshots of how this PR fixes #756:

screenshot of the new "tip distribution" table showing two donations

---

screenshot of the new "tip distribution" table showing one donation at a screen width of 320px

Changaco:
No reviews? 😢

liberapay/security/__init__.py
```diff
@@ -5,48 +5,30 @@ def set_default_security_headers(website, response, request=None):
...
-        if website.canonical_scheme == 'https':
-            csp += b"upgrade-insecure-requests;"
-        response.headers[b'content-security-policy'] = csp
+    csp = (
+        b"default-src 'self' %(main_domain)s;"
+        b"script-src 'self' %(main_domain)s;"
+        b"style-src 'self' %(main_domain)s;"
+        b"connect-src 'self' *.liberapay.org *.mangopay.com *.payline.com;"
+        b"form-action 'self';"
+        b"img-src * blob: data:;"
```
EdOverflow: What about including `object-src`?

Changaco: Why? It's covered by `default-src`. In fact we could even remove `script-src` and `style-src` since they're also identical to `default-src` now.

EdOverflow: Ah, I see you do actually want to set the value of `object-src` to `'self'`. I was going to suggest setting it to `'none'`. In that case, everything looks good to me.

PR #930, opened 2018-01-16
Fix payday

### LINKED ISSUE #909: Payday ignores rate limits and crashes when it hits them
We should modify it to slow down automatically, as a temporary fix while we try to figure out a long term solution (#908).

### LINKED ISSUE #911: Payday doesn't retry interrupted transfers
See <https://github.com/liberapay/salon/issues/194#issuecomment-355015808> and <https://github.com/liberapay/salon/issues/194#issuecomment-355046720>.

Changaco:
Closes #909 and fixes #911.

### DIFF
--- a/liberapay/billing/payday.py
+++ b/liberapay/billing/payday.py
@@ -9,6 +9,7 @@ import os
 import os.path
 from subprocess import Popen
 import sys
+from time import sleep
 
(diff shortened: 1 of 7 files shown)


Changaco:
Rebased on master.

Issue #734, opened 2017-10-13
Redesign


Changaco:
Reticketing from #608 (which will be closed by #731).

This issue is about the possible future redesign of Liberapay's appearance. See <https://github.com/liberapay/liberapay.com/issues/608#issuecomment-323169078> for a design concept.

Bat41:
I really like this design concept. It looks more modern than the actual one. The fonts and colours are also nicer (the yellow we are using is a bit "agressive" IMHO). If everyone is okay with it, I could find some time to work on it.

Changaco:
I would rather not be distracted by a broad visual redesign at this time, I think other issues like #289 and #714 have greater priority.

Changaco:
@Bat41 You seem to dislike the Ubuntu font we're currently using. Can you send me a screenshot of the current homepage on your machine? I'd like to compare the font rendering with mine.

Bat41:
Here is the homepage:

![screen shot 2017-10-13 at 14 21 54-fullpage](https://user-images.githubusercontent.com/16254623/31548310-58ce3a92-b022-11e7-8dcc-d2193b1df404.png)

And a profile page:

![screen shot 2017-10-13 at 14 24 11-fullpage](https://user-images.githubusercontent.com/16254623/31548331-602dbda8-b022-11e7-95b3-6308e7935637.png)

I'm using Open Sans on both.

Changaco:
I meant I want to compare the rendering of the Ubuntu font. This is what the (French) production homepage is supposed to look like:

![liberapay-2017-10-13-homepage-fr](https://user-images.githubusercontent.com/1581590/31549918-e4e78526-b02f-11e7-80f5-ee24cf758055.png)

Bat41:
Ah sorry, I misunderstood what you were asking.

The font seems to be rendered the same on my computer, but in my opinion, Ubuntu is fine only for headings, not paragraphs. But I'm not an expert when it comes to typography, so I may be wrong, it's only my feeling.

PR #935, opened 2018-01-17
Redesign Liberapay

EdOverflow:
This is a PR for #734. My goal is to make Liberapay look less like *bootstrapy* and more unique, and appealing to its users.

image

### DIFF
--- a/www/index.html.spt
+++ b/www/index.html.spt
@@ -73,7 +73,7 @@ recent = query_cache.one("""
     <h2>{{ _("Liberapay is a recurrent donations platform") }}</h2>
     <p>{{ _("We help you fund the creators and projects you appreciate") }}</p>
     <div class="buttons"><a class="btn btn-default btn-lg" href="/sign-up">
-        <span>{{ _("Create an account") }}</span>
+        <span>{{ _("Create an account") }} &rarr;</span>
     </a></div>
     </div>
 
@@ -93,7 +93,7 @@ recent = query_cache.one("""
             <p>{{ _("Are you a creator of commons? Do you make free art, spread free knowledge, write free software?") }}</p>
             <p>{{ _("Yes? Then Liberapay is for you! Create your account, fill your profile, and ask your audience to financially support your work.") }}</p>
             <div class="buttons"><a class="btn btn-success btn-lg" href="/sign-up">
-                {{ glyphicon('log-in') }} <span>{{ _("Create an account") }}</span>
+                <span>{{ _("Create an account") }}</span>
             </a></div>
 
         % else
@@ -199,7 +199,7 @@ recent = query_cache.one("""
             </div>
         </dl>
         <p class="text-center"><br><a class="btn btn-success btn-lg" href="/sign-up">
-            {{ glyphicon('log-in') }} <span>{{ _("Create your account") }}</span>
+            <span>{{ _("Create an account") }}</span>
         </a></p>
     </div>
 
@@ -221,7 +221,7 @@ recent = query_cache.one("""
               >{{ glyphicon('thumbs-up') }} <span>{{ _("Fund Liberapay") }}</span></a></p>
     </div>
 
-    <div class="section">
+    <div class="section text-center">
         <h2>{{ _("Sponsors") }}</h2>
         <p>{{ _(
             "Donations from businesses and nonprofits are welcome on Liberapay. If "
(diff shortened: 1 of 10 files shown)


sirjamesgray:
do you have a before / after image to illustrate the changes?

EdOverflow:
The changes mainly consist of subtle adjustments to individual components of Liberapay. 

| Before | After |
| --------- | -------|
| image | image |

| Before | After |
| --------- | -------|
| image | image |

| Before | After |
| --------- | -------|
| image | image |

| Before | After |
| --------- | -------|
| image | image |

EdOverflow:
@Changaco, when you find time, could you please take a look at the changes I made here? Is there anything you would like me to add or change?

Changaco:
I'm not convinced by some of the proposed changes.

- Dropping the navbar background color makes the website look more "bland" to me, like when it was still gray 2 years ago.
- The footer seems too big (in height).
- The uppercase jumbotron feels like it's shouting at me.

But maybe that's just me. Can we get other opinions?

Also:

- Putting a gray line under links looks okay, but it's a problem for block links (e.g. links that wrap an avatar).
- I remember at least one person speaking up against round avatars in a previous issue.
- It's not okay to make minor changes to translatable strings without updating the PO files.
- This branch needs to be rebased, and we don't use merges inside Pull Requests.
- The first change I would have expected to see in a PR that aims to make Liberapay "look less *bootstrapy* and more unique" is a modification of the theme colors in `variables.scss`, yet this branch doesn't touch them.

LiamDawe:
Just to add my thoughts too, a lot of the changes seem totally unnecessary. Like the navbar, removing the colour? Why? It needs to stand out, it's a navbar.

jorgesumle:
I don't understand much about design, but I recall [@onpon4 didn't like the yellow color at all](https://trisquel.info/de/forum/ive-started-patreon-page#comment-113350) from a forum discussion, here the reasoning:

> That's not what I was talking about. The layout is fine. The problem is the colors, specifically that yellow banner and the yellow "donate" button. This is a terrible color choice and it's even worse to mix it with blue. The text also has bad contrast, so it's difficult to read, and the white background is overly bright.
> Combine these together and if I were to point most people to this page, they would click straight away before I even have a chance to sell it to them.

EdOverflow:
@jorgesumle, that is my opinion on the yellow too. I don't plan on leaving the navbar like this, but I really dislike the current yellow colour on liberapay.com. I will mark this PR as a [WIP] to make it clear to all of you that I am still working on this and by no means done.

comradekingu:
There are no less than 4 buttons to create an account here. I think one is enough.
Donation buttons are more effective in green https://chuffed.org/academy/articles/detail/how-to-increase-your-donation-page-conversion
This also suggests not requiring an account before donating, which may be possible, but is not clear. Waiting for https://github.com/liberapay/liberapay.com/pull/1052 to merge before getting on with rewriting the blocks and the FAQ.

Any amount of design will not overcome the busy look of too much text, which is an improvement directly related to induced stress for the user. I can't seem to overstate this enough.

"Sponsors" and "Thanks" naturally belong at the bottom because it is where one would expect to find them.

MartinDelille:
I personnally prefer this new UI to the current one. Is it still goind on?

I collected the following proposition from the comment:

- [ ] reduce footer height
- [ ] update PO files
- [ ] edit `variables.scss`
- [ ] rebase and fix conflict
- [ ] Remove donate button count to one

www/index.html.spt
```diff
@@ -70,27 +70,39 @@ recent = query_cache.one("""
 
     <div class="jumbotron section">
     <p><img src="{{ website.asset('liberapay/icon-v2_yellow-r.svg') }}" height=100 width=100 /></p>
-    <p>{{ _("Liberapay is a recurrent donations platform.") }}</p>
-    <p>{{ _("We help you fund the creators and projects you appreciate.") }}</p>
+    <h2>{{ _("Liberapay is a recurrent donations platform") }}</h2>
```
comradekingu, line 73:
```suggestion
    <h2>{{ _("Liberapay is a recurrent donation platform") }}</h2>
```

rkingett:
Could we add [AccessLint](https://github.com/marketplace/accesslint) so that our accessibility is constantly reviewed? Especially for this redesign?

The best thing would be to use semantic HTML but the latest Bootstrap has accessibility considerations built in, I think.

Changaco:
I have enabled AccessLint for this repository.

PR #941, opened 2018-01-20
App icon

Changaco:
Adds a link to the 200px yellow icon in the HTML metadata.

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -9,6 +9,7 @@
     <title>{{ full_title }}</title>
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <link rel="icon" href="{{ website.asset('liberapay/icon-v2_black.16.png') }}" type="image/png">
+    <link rel="apple-touch-icon" href="{{ website.asset('liberapay/icon-v2_yellow-r.200.png') }}">
     <link rel="stylesheet" href="{{ website.asset('bootstrap/css/bootstrap.css') }}">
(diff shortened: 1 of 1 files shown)


jorgesumle:
The [microformats page for link rel values](http://microformats.org/wiki/existing-rel-values#HTML5_link_type_extensions) says that the status of this `rel` value is "proposed" and "probably redundant with rel=icon". I didn't found it in the HTML 5.2 Recommendation nor in the [HTML
Living Standard](https://html.spec.whatwg.org/multipage/links.html#other-link-types). 

You can specify multiple icon sizes using the recommended [rel=icon](https://www.w3.org/TR/html5/links.html#link-type-icon), by specifying a `size` attribute. There is even an example there similar to what you're trying to do: `<link rel=icon href=iphone.png sizes="57x57" type="image/png">`.

Changaco:
Unfortunately `apple-touch-icon` is the *de facto* standard for now. The problem with using `icon` is that, like many other websites, we use one style for the small icon (simple black), and another for the large icon (yellow background), but the `sizes` attribute isn't well-supported enough by browsers yet.

jorgesumle:
The solution is making the free software programs use the standard approach. Firefox (and its derivatives) doesn't use the standard approach in the favorite pages menu (instead it uses the apple non-standard rel attribute).

In the meantime, we can use this *de facto* standard, which names the apple brand. It's just an HTML tag, so I'm OK with the merge.

PR #955, opened 2018-01-26
Improve i18n subdomains

### LINKED ISSUE #927: New languages fy and et don't work
The new languages [fy](https://fy.liberapay.com/) and [et](https://et.liberapay.com/) don't seem to work. It doesn't load any page on those subdomains.

Changaco:
Closes #927 and attempts to reduce traffic from crawlers.

### DIFF
--- a/liberapay/utils/__init__.py
+++ b/liberapay/utils/__init__.py
@@ -14,6 +14,7 @@ from operator import getitem
 import os
 import pickle
 import re
+import socket
 
 from six import PY3
 from six.moves.urllib.parse import quote as urlquote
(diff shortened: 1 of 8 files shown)


Changaco:
I went further and removed the navbar language dropdown from profile pages. If the user doesn't have a statement in a specific language then linking to that language's subdomain is probably not useful.

Changaco:
I've removed from Cloudflare the 6 existing language subdomains which were less than 50% translated.

Changaco:
(Greek, Turkish, Japanese, Indonesian, Hungarian, and Ukrainian.)

PR #960, opened 2018-01-28
Improve handling of saved cards

### LINKED ISSUE #681: Card payments with saved number lead to confusing failures
We allow users to save their card number so they don't have to input it again next time, however this is currently resulting in payment failures with a confusingly generic error message (`009499: Bank technical error`). What this apparently means is that the payer's bank is refusing the transaction because the [card's verification number](https://en.wikipedia.org/wiki/Card_security_code) is missing. Why is it missing? Because payment processors are not allowed to store it.

MangoPay is working on a solution to this problem. In the meantime we should either stop saving card numbers, or ask the user to re-input their card data when we get 009499 error code.

Changaco:
Fixes #681.

### DIFF
--- a/www/%username/wallet/payin/card/%back_to.spt
+++ b/www/%username/wallet/payin/card/%back_to.spt
@@ -51,16 +51,6 @@ else:
     route = routes[0] if routes else None
     del routes
 
-card = Card.get(route.address) if route else None
-if card:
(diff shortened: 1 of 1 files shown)


Changaco:
screenshot_20180128_144842

Issue #986, opened 2018-02-22
Password check


Changaco:
When a user inputs a password we should check if it's in the "pwned passwords" database: [I've Just Launched "Pwned Passwords" V2 With Half a Billion Passwords for Download](https://www.troyhunt.com/ive-just-launched-pwned-passwords-version-2/).

umang-malik:
I would like to try this. What shall I do after checking if it exists in the database? Maybe set a threshold value for the rarity of the password?

Changaco:
Hey @umang-malik, welcome.

Given that Liberapay is a service that handles money we should show a warning no matter what the password count is. However the wording of the warning could vary based on the password count, and maybe the password's "strength" too.

We need to take into account that there can be false positives (hash collisions). For example, if looking up the hash of a complex password returns a high leak count, then it may be a hash collision with a weak password.

umang-malik:
@Changaco, please review #988. I believe that taking hash collisions into considerations while setting a new password is unneecessary. The thresholds have been decided by checking various passwords.

PR #988, opened 2018-02-26
Password check

umang-malik:
Whenever the user sets/changes his password, his password is checked in the "pwned passwords" database (https://www.troyhunt.com/ive-just-launched-pwned-passwords-version-2/). The strength of the password based on the characters is also checked.
 If the password is found weak on the basis of the above, then the user is suggested to use a better password.
This PR is with reference to #986.


umang-malik:
1. Ask user for confirmation using `response.render('templates/confirm.spt', ...)`. Using `out['msg']` for successful change in password.
2. Changed to `/range` endpoint of pwnedpasswords instead of sending the entire hash.

@Changaco, I don't understand why we need to check the strength of the password during log-in. The password has already been assessed while setting/changing it.
Also, I am not sure why we need to store the time of last check in our DB. There is already a restriction upon changing passwords, which itself restricts the API querying.

Changaco:
We need to check during log-in because:

1. most people never change their password
2. we already have more than 3400 passwords in our database
3. the pwnedpasswords database can grow, a password that isn't in it now may end up in it later

www/%username/settings/edit.spt
```diff
@@ -25,8 +28,34 @@ if 'new-password' in body:
...
+            if passhash_short + suffix == passhash:
+                count = int(line.split(":")[1].strip())
+        strength, improvements = passwordmeter.test(body['new-password'])
+
+        back_to = p.path('settings')
+        if strength > 0.3 and count < 2000 :
+            msg = _(
+                "Are you sure you want to change your password?"
+            )
+            raise response.render('templates/confirm.spt', state, cls='warning', msg=msg, back_to=back_to)
```
Changaco: This confirmation warning is unnecessary, we should only warn the user when the password is weak or compromised.

www/%username/settings/edit.spt
```diff
@@ -25,8 +28,34 @@ if 'new-password' in body:
...
+        else:
+            msg = _(
+                "The new password is not strong enough. We recommend you to choose a combination of letters, symbols, and numbers."
+                "Are you sure you want to change your password?"
+            )
+            raise response.render('templates/confirm.spt', state, msg=msg, back_to=back_to)
```
Changaco: We need at least two different warning messages here, one when the password is weak, and another when the password is compromised (i.e. when it was found in the pwned database).

www/%username/settings/edit.spt
```diff
@@ -25,8 +28,34 @@ if 'new-password' in body:
...
     p.update_password(body['new-password'])
-    out['msg'] = _("Your password has been changed.")
+    out['msg'] = _("Your password has been changed")
```
Changaco: This change should be reverted.

Changaco:
To keep track of password checks I suggest using the `add_event` and `get_last_event_of_type` methods, with a new `'password_check'` event type and an empty payload (because all we need is a timestamp, and that's the event time `ts`).

umang-malik:
Where should I implement the login-check? I am currently doing it in `liberapay/security/authentication.py` but am unable to send the notification to the user once the user logs in.

umang-malik:
@Changaco, [`p.add_event(website.db.get_cursor(), 'password-check', None) `](https://github.com/liberapay/liberapay.com/pull/988/files#diff-4a6e6cdf98c11ac6672aca20fdf78f22R86) line is giving an error which I am unable to fix: 
```Traceback (most recent call last):
  File "/home/umang/Desktop/folder/liberapay.com/env/lib/python2.7/site-packages/algorithm.py", line 321, in loop
    new_state = function(**deps.as_kwargs)
  File "/home/umang/Desktop/folder/liberapay.com/liberapay/security/authentication.py", line 195, in authenticate_user_if_possible
    p = sign_in_with_form_data(body, state)
  File "/home/umang/Desktop/folder/liberapay.com/liberapay/security/authentication.py", line 88, in sign_in_with_form_data
    p.add_event(website.db.get_cursor(), 'password-check', None)
  File "/home/umang/Desktop/folder/liberapay.com/liberapay/models/participant.py", line 1126, in add_event
    return c.one("""
AttributeError: 'CursorContextManager' object has no attribute 'one'
```
Also, how do I notify the user about his weak password? I tried using `%if state.get('password-check')` in `templates/log-in-form.html` but it is not working.

Changaco:
@umang-malik Dropping `.get_cursor()` should work, i.e.: `p.add_event(website.db, 'password-check', None)`.

To notify the user I see two possibilities:

- create a new `templates/password-warning.spt` template and call `response.render(...)`
- create a new `emails/password-warning.spt` template and call `p.notify(...)`

The latter should be simpler to implement, and it would allow moving the password check to a background worker, so I think it's the best option.

umang-malik:
@Changaco, All the requested things have been done. :smiley:
The time period for password check during login has been set to 30 days for now. Also, I have currently disabled an email notification for the same.
Please review the PR.

emails/password_warning.spt
```diff
@@ -0,0 +1,5 @@
+{{ _("Update your Liberapay password") }}
+
+[---] text/html
+<p>{{ _("{0}",message) }}</p>
```
Changaco: This isn't right, `_("{0}", message)` is a no-op. You shouldn't send an already-translated message to a notification template. Instead you should send the variables that the template needs to generate the message. In this case it could be either `password_strength` and `password_leak_count`, or a `password_status` string (possible values: `'weak'`, `'common'`, `'compromised'`). The latter is probably better because it allows changing the thresholds without breaking old notifications.

liberapay/security/authentication.py
```diff
@@ -14,6 +17,9 @@
 from liberapay.models.participant import Participant
 from liberapay.utils import get_ip_net
 
+import passwordmeter
+from hashlib import sha1
+import requests
```
Changaco: Please don't break the import style like this. Imports are split into specific groups (based on [PEP8](https://www.python.org/dev/peps/pep-0008/#imports)), and ordered alphabetically within each group.

liberapay/security/authentication.py
```diff
@@ -57,6 +63,41 @@ def sign_in_with_form_data(body, state):
...
+                        )
+                        p.notify('password_warning', email=False, message=password_check_message)
+                    elif count > 0:
+                        password_check_message = _(
+                            "We have detected that your current password has been compromised in leaks "
+                            "on the internet before. "
+                            "You are recommended to change your password."
+                        )
+                        p.notify('password_warning', email=False, message=password_check_message)
+                    p.add_event(website.db, 'password-check', None)
```
Changaco: This should be moved to a separate function, and the call to that new function should be wrapped to report and mask errors, like this:

```python
try:
    check_password(password)
except Exception as e:
    website.tell_sentry(e, state)
```

liberapay/security/authentication.py
```diff
@@ -57,6 +63,41 @@ def sign_in_with_form_data(body, state):
             )
             if not p:
                 state['log-in.error'] = _("Bad username or password.")
+            else:
+                last_password_check = p.get_last_event_of_type('password-check')
+                if (not last_password_check) or utcnow() - last_password_check.ts > timedelta(days=30):
```
Changaco: I don't think we need to check every 30 days, once or twice per year should be enough, e.g. `days=180`.

www/%username/settings/edit.spt
```diff
@@ -25,8 +28,47 @@ if 'new-password' in body:
...
+        passhash_short = passhash[:5]
+        URL = "https://api.pwnedpasswords.com/range/"+passhash_short
+
+        r = requests.get(url = URL)
+        count = 0
+        for line in r.text.split("\n"):
+            suffix = line.split(":")[0]
+            if passhash_short + suffix == passhash:
+                count = int(line.split(":")[1].strip())
+        strength, improvements = passwordmeter.test(body['new-password'])
```
Changaco: This is code duplication, it should be replaced with a call to a separate `check_password` function.

requirements_base.txt
```diff
@@ -173,3 +173,13 @@ ipaddress==1.0.18 \
...
+
+asset==0.6.12 \
+    --hash=sha256:addb25dbdbfa820c667e246f9c09122401df21cd9e94db919c72f8f04fc972eb
+passwordmeter==0.1.8 \
+    --hash=sha256:f4f7df54182072a137bca3e396f2dad354770ba596dd41f6646cd1f0665fee12
+aadict==0.2.3 \
+    --hash=sha256:a77328ac55dbb5735da99441870251befe135f687ab707a7a178561363b27704
+globre==0.1.5 \
+    --hash=sha256:ee214204f237e9114b8f61eeb61c2abd1e665ca3b59e5a6a0b070971c0bb12e2
+
```
Changaco: I've looked through these new dependencies, they appear to be okay (although their source code could use a good cleanup).

One issue is that `passwordmeter` includes [its own list of ten thousand common passwords](https://github.com/cadithealth/passwordmeter/tree/v0.1.8/passwordmeter/res), and [always loads it](https://github.com/cadithealth/passwordmeter/blob/v0.1.8/passwordmeter/__init__.py#L25-L26), even if we skip the `notword` factor (which we should do since it's redundant with the pwnedpasswords check). This wastes approximately 1.2MB of RAM (measured with [this `get_size` function](https://stackoverflow.com/a/40880923/2729778) on my laptop).

umang-malik:
@Changaco , all the changes have been made.

- I have added a seperate method to check password.
- The notword factor in passwordmeter has been skipped. I can't think of an easy solution to the extra memory required.
- `tests/py/fixtures/TestLogIn.yml` contains the details of the requests made to the pwnedpasswords API. The response headers change every time the tests are run, and hence adding the file to `.gitignore` seems appropriate.

Changaco:
I've made lots of changes.

@umang-malik You should look at them and let me know what you think.

Screenshots:

---

screenshot of a warning displayed when a user attempts to protect their account with a weak common password

---

screenshot of a warning notification shown after a user logged in with a weak common password

Changaco:
I dropped the password strength test because it wasn't good. See [this blog post](https://blogs.dropbox.com/tech/2012/04/zxcvbn-realistic-password-strength-estimation/) for a complete explanation.

I didn't replace `passwordmeter` with [`zxcvbn-python`](https://github.com/dwolfhub/zxcvbn-python) because its algorithms are language-specific and by default only English is supported. We can try to integrate `zxcvbn` in the future, but it's not a priority.

umang-malik:
@Changaco, I agree with the given blog's results.Though I think that keeping a measure of password strength would help in setting a minimum threshold for passwords, especially in languages other than English, as the pwnedpassword database is heavily dominated by the it (and other common languages).

For example, `hello1234` appears ~20,000 times, as compared to 172 of `bonjour1234`, or that `müller` just appears 11 times! Some common passwords in less-common languages might pass the pwnedpasswords test.
Other than that, the pwnedpasswords test looks good enough.

umang-malik:
Thanks for improving the templates. They look a lot better than mine.

Changaco:
The truth is that for Liberapay anything that's not a long string of random characters is a bad password. It doesn't really matter whether it's `hello1234` or `correcthorsebatterystaple`, or how secure the latter is compared to the former, because [the only secure password is the one you can’t remember](https://www.troyhunt.com/only-secure-password-is-one-you-cant/).

umang-malik:
Well, I guess it shall be fine. We can always pull up a new issue in case we feel the need to do so. 
P.S. All this reading about passwords has made me realize how much I need to work on my own passwords. :sweat_smile:

Changaco:
Rebased on master and cleaned up the commit messages.

Changaco:
@umang-malik In the future you can drop the `Signed-off-by` lines in commit messages.

Changaco:
Deployed and [announced](https://github.com/liberapay/salon/issues/208#issuecomment-370424494).

Changaco:
I've emailed Troy Hunt to let him know about this.

PR #1007, opened 2018-03-09
Improve money amount input

### LINKED ISSUE #786: Donation should be number pad on mobile and text area should be number

Changaco:
Use `<input type="tel" inputmode="decimal">` for money amounts.

([`inputmode`](https://html.spec.whatwg.org/multipage/interaction.html#attr-inputmode) isn't actually implemented in any browser, but maybe someday it will be.)

We can't use `type="number"` because it's not handled properly by some browsers (see the red cells of the table in <https://www.ctrl.blog/entry/html5-input-number-localization>). I tested Chromium on Linux, and even with both the UI and the page being in French the browser rejects commas.

Closes #786. Supersedes #848.


Changaco:
In case future browser versions fix the bugs mentioned above, note that it's possible to hide the "arrows" that browsers show for `type="number"` using this style:

```css
input[type="number"] {
    -moz-appearance: textfield;
    -webkit-appearance: textfield;
    appearance: none;
}
```

Links:

- https://developer.mozilla.org/en-US/docs/Web/CSS/-moz-appearance
- https://drafts.csswg.org/css-ui-4/#appearance-switching
- https://css-tricks.com/almanac/properties/a/appearance/
- https://caniuse.com/#feat=css-appearance

Changaco:
Other relevant links:

- http://mobileinputtypes.com/
- http://blog.pamelafox.org/2012/05/triggering-numeric-keyboards-with-html5.html
- https://stackoverflow.com/a/31619707/2729778

jorgesumle:
The `tel` attribute is for telephone numbers not for regular numbers. This may work for most phones but it's not correct HTML.

Source: https://www.w3.org/TR/html/sec-forms.html#telephone-state-typetel

Changaco:
The `tel` type is *de facto* for any free-form numeric input. In fact we should probably use it for other digital inputs like card numbers and postal codes.

PR #1013, opened 2018-03-16
Multiple currencies for teams - part 2

Changaco:
Closes #982.

PR #1016, opened 2018-03-18
Improve currency selector for team takes

Changaco:
While writing the announcement for #1013 I realized that a detail of the implementation didn't make sense. The new commit restricts the choice of currencies to those accepted by both the member and the team.


Changaco:
> The new commit restricts the choice of currencies to those accepted by both the member and the team.

That didn't make sense either, so I've switched to a choice between the team's main currency and the member's main currency, with the current currency of the take as an extra "choice" to cover takes that were created at a time when the currency settings were different.

PR #1041, opened 2018-04-01
fix #1018: ctrl-clicking project names "new tab" behavior

### LINKED ISSUE #1018: Control clicking project name opens link in current and new tab
When control clicking a project name on https://liberapay.com/explore/teams I expect the browser to open the link in only a new tab, yet it opens in both a new tab and the current tab.  When I right click and select open in new tab, the browser only opens a new tab.  Tested on Firefox 61 nightly.

hpr:
removing these lines fixed the issue on my self-hosted install, and didn't fail any tests in `make test`.

### DIFF
--- a/js/10-base.js
+++ b/js/10-base.js
@@ -40,7 +40,8 @@ Liberapay.init = function() {
     Liberapay.s3_uploader_init();
 
     $('div[href]').css('cursor', 'pointer').click(function() {
-        window.location.assign(this.getAttribute('href'));
+        location.href = this.getAttribute('href');
+        return false;
     });
     
     $('.navbar .dropdown-hover').removeClass('dropdown-hover');
--- a/js/10-base.js
+++ b/js/10-base.js
@@ -39,10 +39,6 @@ Liberapay.init = function() {
     Liberapay.payments.init();
     Liberapay.s3_uploader_init();
 
-    $('div[href]').css('cursor', 'pointer').click(function() {
-        location.href = this.getAttribute('href');
-    });
-
     $('.navbar .dropdown-hover').removeClass('dropdown-hover');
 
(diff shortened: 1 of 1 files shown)


Changaco:
Removing the ability to click on the entire `<div>` is not acceptable. We should either improve the JS code, or switch to using the `overlay` and `overlayed` CSS classes.

hpr:
@Changaco Thanks. Doing `window.location.assign(this.getAttribute('href'));` instead looks like at least an improvement on the current situation, because it still allows you to click on the entire `<div>` and ctrl-clicking on the link part of the div works as expected, though ctrl-clicking on non-link parts of the div does nothing. I'll look into a better way, so that ctrl-clicking on the entire div works as expected.

hpr:
I pushed that change, and it looks like the best solution I can think of -- the only other alternative would be doing something like this, which is a hack and also would not work with middle-click, cmd-click on Mac, etc: https://stackoverflow.com/a/18639931/6632587

hpr:
From talking with [/u/kozec on reddit](https://www.reddit.com/r/linux/comments/8aixg0/a_new_version_of_flathub_to_browse_flatpak_apps/dwzmsi7/) I realized that returning `false` from the onclick function fixed the duplication bug, though it still opens in the same tab with ctrl-click (middle-click works fine, though). 

This new pushed change is just the original code but returning `false` at the end, and ctrl-clicking on this new link does not open two copies of the same page for me on Firefox ESR. As it fixes the original bug and maintains the ability to click on the entire `div`, I think this is good enough to merge for the time being. Thanks for your patience!

hpr:
@jorgesumle Sure, that was my bad. Done!

Changaco:
Good work @hpr!

PR #1043, opened 2018-04-01
Support Libravatar Federation

criztovyl:
By using pyLibravatar this PR allows Liberapay to support Libravatar federation.

I had no time for cloning the repo, so I did the changes online using the GitHub editor.

I tested this locally on my notebook, there Python hangs while doing the required DNS request, b/c no TCP DNS is available.
I also tested it on my Webspace, there it works.

### DIFF
--- a/liberapay/models/participant.py
+++ b/liberapay/models/participant.py
@@ -1,6 +1,5 @@
 from __future__ import division, print_function, unicode_literals
 
-from libravatar import libravatar_url
 from base64 import b64decode, b64encode
 from datetime import timedelta
 from email.utils import formataddr
@@ -14,6 +13,7 @@ from six.moves.urllib.parse import quote, urlencode
 import aspen_jinja2_renderer
 from cached_property import cached_property
 from html2text import html2text
+from libravatar import libravatar_url
 import mangopay
 from mangopay.utils import Money
 from markupsafe import escape as htmlescape
--- a/liberapay/models/participant.py
+++ b/liberapay/models/participant.py
@@ -1567,7 +1567,7 @@ class Participant(Model, MixinTeam):
         if platform == 'libravatar' or platform is None and email:
             if not email:
                 return
-            avatar_url = libravatar_url(email)
+            avatar_url = libravatar_url(email, https=True)
             avatar_url += AVATAR_QUERY
 
         elif platform is None:
--- a/liberapay/models/participant.py
+++ b/liberapay/models/participant.py
@@ -1,5 +1,6 @@
 from __future__ import division, print_function, unicode_literals
 
+from libravatar import libravatar_url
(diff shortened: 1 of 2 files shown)


criztovyl:
So far, so good.

One remaining question from my side would be if there is a better way to check whether the website is https other than assuming it?

Thank you.

Changaco:
This is a bad idea, the cost is high while the benefit is near-zero.

At the very least the call to `libravatar_url()` needs to be wrapped to catch exceptions and fall back to the `libravatar.org` URL we currently use. It also needs to time out automatically if the DNS request takes longer than 1 second.

criztovyl:
I understand. I will try fix that :)

criztovyl:
Some thought from myself, from what I realized now.

The method I changed is called *each time* the page is rendered, so I consider my change as bad now. The full round-trip that hides behind the method call (mail -> SRV -> host) is impossible at such a position.

What I would consider to be better would be that the round-trip is only be done once after the user changes it's mail address, storing the result (the avatar hostname*) in the user's profile. (I do not know how profile data is handled currently.) The underlying SRV record might change, so the saved result might need to be refreshed after the SRV record's TTL is reached (Libravatar requires that the TTL must be >= 1 day).

Would that lower the cost enough being worth it?

---
\* The hostname stored in the SRV record should be enough as the other parameters are either fixed (the /avatar/ sub-path) or can be determined (hashing the mail address).

Changaco:
The cost I was referring to is the time needed to get this merged and then maintain it.

The `update_avatar()` method is only called when something has changed, not "each time the page is rendered".

Changaco:
Libravatar is shutting down: #1044.

Changaco:
I think we should add support for [WebFinger](https://webfinger.net/) ([RFC 7033](https://tools.ietf.org/html/rfc7033)) instead of the Libravatar protocol.

Changaco:
Basically send an HTTP request for `https://<email domain>/.well-known/webfinger?resource=acct:<url-encoded email address>&rel=http://webfinger.net/rel/avatar` and look in the returned JSON for a link whose `rel` attribute is `http://webfinger.net/rel/avatar`.

tastytea:
[Libravatar is not shutting down anymore](https://blog.libravatar.org/posts/Libravatar.org_is_not_going_away/).

@criztovyl Do you intent to continue working on this PR?

criztovyl:
Not at the moment, no. (Time is unfortunately spare currently.)

I would say the approach this PR takes is also not worth extending it, a new PR would be better. (Its only 10 changes anyways.) Will close this PR and open a feature request instead.

Regarding the webfinger idea, I think its worth it's own feature request with an issue of it's own, not buried in here. I will open one.

PR #1045, opened 2018-04-04
Fix updating of elsewhere data

### LINKED ISSUE #1042: The `refetch_repos` cron is broken
It stopped working more than 5 months ago. It's stuck on a GitHub account that no longer exists (or has been renamed).

Changaco:
Closes #1042. Also solves #836, but at the cost of losing support for getting the list of channels a user follows.


Changaco:
I've removed all the Twitch stream keys we had in our database (`update elsewhere set extra_info = extra_info::jsonb - 'stream_key' where platform = 'twitch'`).

PR #1053, opened 2018-04-09
Fix updating of elsewhere data - part 2

Changaco:
Continuation of #1045.

### DIFF
--- a/liberapay/elsewhere/_base.py
+++ b/liberapay/elsewhere/_base.py
@@ -259,7 +259,7 @@ class Platform(object):
         members = [self.extract_user_info(m, domain) for m in members]
         return members, count, pages_urls
 
-    def get_user_info(self, domain, key, value, sess=None):
+    def get_user_info(self, domain, key, value, sess=None, uncertain=True):
         """Given a user_name or user_id, get the user's info from the API.
         """
         if key == 'user_id':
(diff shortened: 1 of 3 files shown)


liberapay/elsewhere/_base.py
```diff
@@ -144,9 +144,10 @@ def api_get(self, domain, path, sess=None, **kw):
         return self.api_request('GET', domain, path, sess=sess, **kw)
 
     def api_error_handler(self, response, is_user_session, domain):
+        response_text = response.text  # for Sentry
```
Changaco: screenshot_20180409_143620

Changaco: I've added `response_text` to the "safe" list:

screenshot_20180409_145122

Changaco:
Since mastodon.rocks lost all its data I've deleted it in our DB as well:

```sql
delete from elsewhere where domain = 'mastodon.rocks';
DELETE 23
delete from oauth_apps where domain = 'mastodon.rocks';
DELETE 1
```

Changaco:
We're also getting lots of `InvalidGrantError`s from Google. There are many possible causes, here's a good post about that: [Google OAuth “invalid_grant” nightmare — and how to fix it](https://blog.timekit.io/google-oauth-invalid-grant-nightmare-and-how-to-fix-it-9f4efaf1da35). In our case we'll just treat this error the same as a `TokenExpiredError`.

Issue #261, opened 2016-04-09
Implement 'keep me logged in'


revi:
I'd like to be kept logged in for a while (ie. 2-4 weeks) so I don't have to log in often.

Maybe we need to implement 2FA first for re-authenticating for sensitive operations.

Changaco:
2FA is not required, we only need to prompt the user to reenter their password before making important changes, such as modifying donation amounts.

tuxayo:
Some keywords to make this issue more easy to find (some of the things I tried before finally finding this)
- logout
- logged out
- disconnected
- disconnect
- session

So from what I understand it's normal to currently have to log in something like once per day?

Changaco:
> So from what I understand it's normal to currently have to log in something like once per day?

Yes.

tuxayo:
Good, at least it's not some strange bug :)

Should the issue be split? There is the session duration topic and the 2FA topic.

Changaco:
The issue for 2FA is #926.

PR #1069, opened 2018-04-12
Improve handling of "user secrets"

Changaco:
This branch cleans up and optimizes our handling of user passwords and sessions. The goal is to reduce the amount of data transferred between the python app and the database, by moving the secrets (which are long strings accompanied by timestamps) out of the `participants` table. Doing this also reduces the risk of accidental leaks of user secrets since we're only fetching them from the database when we actually need them. Finally, this opens the door to having multiple sessions per user, which is pretty much a requirement for #261.


Changaco:
Oh, this branch also removes obsolete support for HTTP Basic authentication, and increases the strength of session passwords (from less than 128 bits to 192 bits) without increasing the cost (thanks to a switch from hexadecimal to base 64).

Changaco:
Deployment done. No breakage so far.

Issue #1073, opened 2018-04-16
Language variants


Changaco:
Our current i18n infrastructure only supports "top-level" languages, it would be great if it also supported "variants" like Brazilian Portuguese, without adding thousands of duplicate lines in the `i18n/` directory.

RockyTV:
Is it hard to change the current code to support variants?

jorgesumle:
Apache redirections will be needed to redirect the currently translated `pt.liberapay.com` to `pt_PT.liberapay.com`, the same with Spanish and other translated languages. I'm worried that the URLs look weird.

Changaco:
@RockyTV It's hard to do it without getting thousands of duplicate lines.

@jorgesumle We won't redirect anything, and we probably shouldn't create subdomains like `pt-br` or `br.pt`.

RockyTV:
Why not use [ISO 3166-2](https://en.m.wikipedia.org/wiki/ISO_3166-1_alpha-2) country codes for subdomains?

RockyTV:
Okay, I don't know how it works at the moment, but couldn't you just use the `Accept-Language` HTTP header and set a cookie that defines which language you'd like to use? And if you want to change your language, it will just fetch the value from the cookie, allowing the user to explicitly define which language they want to browse the site in.

Changaco:
We already use the `Accept-Language` header.

The subdomains provide public URLs for each version of a page, which allows linking to a specific version and enables proper indexing by search engines.

This issue is not about cookies or subdomains, it's about sharing translations between language variants. The file format we currently use to store translations doesn't support that.

PR #1074, opened 2018-04-18
Optimize memory consumption of i18n module

Changaco:
Yesterday as I was thinking about sharing translations between language variants (#1073) I realized that we were probably not even sharing source strings between catalogs in RAM, and I was right. This branch will fix that, thus reducing memory usage by several megabytes.


Changaco:
Upstream issue: <https://github.com/python-babel/babel/issues/571>.

PR #1111, opened 2018-05-22
Add badges to widgets page

### LINKED ISSUE #1055: Add shields.io badges to /widgets
Following deployment of #775, we should let people know these exist. I was reminded of this by https://twitter.com/ELWillis10/status/983356205887315968.

mattbk:
Closes #1055.

This includes the logo even though the logo [has not been deployed yet](https://github.com/badges/shields/pull/1692) at shields.io.

image

Ready for review.

### DIFF
--- a/www/%username/widgets/index.html.spt
+++ b/www/%username/widgets/index.html.spt
@@ -68,9 +68,11 @@ response.headers[b'Content-Security-Policy'] = csp
     <p><script src="giving.js"></script></p>
     % endif
 
-    <h3>{{ _("Badges from {0}Shields.io{1}", 
-        "<a href='https://shields.io'>"|safe,
-        "</a>"|safe) }}</h3>
+    <h3>{{ _(
+        "Badges from {link_open}Shields.io{link_close}",
+        link_open="<a href='https://shields.io'>"|safe,
+        link_close="</a>"|safe
+    ) }}</h3>
 
     <p>{{ _("Use these code snippets to add a badge to your website or README:") }}</p>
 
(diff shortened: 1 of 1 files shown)


www/%username/widgets/index.html.spt
```diff
@@ -68,5 +68,25 @@ response.headers[b'Content-Security-Policy'] = csp
     <p><script src="giving.js"></script></p>
     % endif
 
+    <h3>{{ _("Shields.io Badges") }}</h3>
+
+    <p>{{ _("Use this code to add a badge to your website or GitHub:") }}</p>
```
Changaco: I see a couple of issues with this wording:

- "this code" is singular, but there are 4 different pieces of code
- "GitHub" is only one platform, I think "README" would be more generic and more precise at the same time (it says exactly where the badges are meant to be included)

www/%username/widgets/index.html.spt
```diff
@@ -68,5 +68,25 @@ response.headers[b'Content-Security-Policy'] = csp
     <p><script src="giving.js"></script></p>
     % endif
 
+    <h3>{{ _("Shields.io Badges") }}</h3>
```
Changaco: It might not be immediately clear to translators and users that "Shields.io" is a domain name here. Maybe "Badges (shields.io)" or "Badges from shields.io" would be better?

Changaco:
> This includes the logo even though the logo has not been deployed yet at shields.io.

Do you think we should wait for it to be deployed?

www/%username/widgets/index.html.spt
mattbk: Addressed in https://github.com/liberapay/liberapay.com/pull/1111/commits/a3e34cef4e908b5f252d5d533fd3c562f0cdcd54. "Shields.io" is capitalized because that seems to be the way they write it.

mattbk:
> Do you think we should wait for it to be deployed?

It may be confusing to see `logo` on the URL but no logo on the badges themselves. So we can either wait until that is deployed or modify the URLs for now.

PR #1114, opened 2018-05-24
Wording in the FAQ

trebmuh:
It sounds unusual when using a singular form in the first part, then a plural form in the second part.
Suggestion is to keep it consistent across the sentence.
Then there are 2 possibilities:
1) all singular,
2) all plural.

I choose the singular form to keep it light.

### DIFF
--- a/www/about/faq.spt
+++ b/www/about/faq.spt
@@ -72,8 +72,8 @@ title = _("Frequently Asked Questions")
     <dd>{{ _(
         "You can usually check the authenticity of a Liberapay profile by looking at "
         "the social accounts connected to it. When a user connects an external account "
-        "(for example a {platform1} or {platform2} account) to a Liberapay profile they "
-        "automatically prove that they control these accounts. You can also look for a "
+        "(for example a {platform1} or {platform2} account) to a Liberapay profile he "
+        "automatically proves that he controls this accounts. You can also look for a "
         "link to a Liberapay profile in a project's official website. Finally if you "
         "have doubts about a specific account you can ask us and we'll look into it."
         , platform1=website.platforms.list[0].display_name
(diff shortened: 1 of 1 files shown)


Changaco:
"they" is not plural here, it's a gender-neutral singular.

Changaco:
https://en.wiktionary.org/wiki/they

trebmuh:
Quoting from wikipedia:

> usually plural, sometimes singular

It's confusing. My guestimate is that most people (specially the non-English native like me) are expecting they to reflect plural.

trebmuh:
Alternate wording, less confusing IMHO:

> When a user connects an external account (for example a {platform1} or {platform2} account) to a Liberapay profile, that automatically proves that this user controls this account.

mattbk:
The alternate wording works for me. Conversely, you could write something like
> Control of a Liberapay account is automatically proven by connection of one or more external accounts (for example a {platform1} or {platform2} account).

jorgesumle:
@trebmuh, I was confused until I learnt about it.

> It's confusing. My guestimate is that most people (specially the non-English native like me) are expecting they to reflect plural.

I wouldn't change it unless native English-speakers would complain about it. @mattbk, does it look correct to you?

mattbk:
I'm a traditionalist in that I expect "they" to be plural, so I may not be much help. It jumps out to me, but someone younger than I or who has less experience with formal writing may not even notice the issue being pointed out here.

trebmuh:
@Changaco @mattbk @jorgesumle : I was pretty much out of the web with a full-time training on the last weeks. I've PR @mattbk 's proposition which sounds clearer to me.

If you do agree, please merge.

jorgesumle:
I think that you need to put a comma after "for example".

trebmuh:
Good catch. Done.

Changaco:
@jorgesumle You should not have merged this, and especially not with a merge commit. An atomic 3-line change should be a single commit, not 5!

PR #1133, opened 2018-06-05
Restrict cookies to same-site requests by default

### LINKED ISSUE #1100: Restrict cookies to same-site requests
[Support for Same-Site cookies has landed in Firefox 60](https://blog.mozilla.org/security/2018/04/24/same-site-cookies-in-firefox-60/). We should add `SameSite=lax` to all our cookies as an additional protection against CSRF. Bonus: this should speed up widgets included in third-party websites, and enforce privacy by preventing the tracking of users through those widgets.

Changaco:
Closes #1100.

### DIFF
--- a/liberapay/main.py
+++ b/liberapay/main.py
@@ -171,6 +171,14 @@ algorithm.functions = [
 ]
 
 
+# Monkey patch python's stdlib
+# ============================
+
+from six.moves.http_cookies import Morsel
+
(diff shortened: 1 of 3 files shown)


Changaco:
(I've shared the monkey-patch used in this PR in <https://stackoverflow.com/q/50813091/2729778>.)

PR #1149, opened 2018-06-19
Fix typo in german translation

Art4:


Art4:
Fixed the typo on weblate.

PR #1153, opened 2018-06-21
Remove trailing whitespace from statement title

jorgesumle:


Changaco:
Why?

jorgesumle:
screenshot at 2018-06-21 13 54 57

It's inconvenient when you copy and paste, it looks ugly and it's also useless.

Changaco:
We probably have quite a few extra spaces like this one, I don't think trying to eliminate them one by one using Jinja's whitespace remover is a good idea, especially since they're a minor inconvenience that most users will never notice.

Related issue: #80.

jorgesumle:
You're right, there are quite a few more. I didn't find another better way to remove them with Jinja.

PR #1163, opened 2018-07-02
Implement self-service card payin refunds

Changaco:
Part of the preparation for the upcoming mangopay exodus.


Changaco:
Explanation of why direct debits are excluded:

Screenshot of the mangopay documentation. Text: "You should be very careful when refunding these direct debit payments because chargebacks (disputes) are non-contestable, therefore if the end-user also requests a chargeback to their bank for a payment you have already refunded, you will have to cover the disputed funds yourself (since the funds of the original payment have already been refunded)."

Source: <https://docs.mangopay.com/endpoints/v2.01/payins#e289_the-direct-debit-direct-payin-object>.

Changaco:
This branch is in need of a few tests.

PR #1167, opened 2018-07-06
update Mastodon account_url to better support Pleroma

mabgnu:
Unlike Mastodon, Pleroma doesn't support the {domain}/@{user_name} route,
not at the moment at least. However, they both support {domain}/users/{user_name}.


mabgnu:
I guess this is going nowhere.

mattbk:
I think this is a good idea, but perhaps some more work could be done to let users know that either Pleroma or Mastodon will work here. What about other Fediverse platforms?

trebmuh:
Pinging @Changaco . Any thought / recommendation here ?

Issue #959, opened 2018-01-27
Commas with Conditional Sentences


jorgesumle:
I've seen a bunch of sentences that don't have a comma after the dependent clause (`if`); they should. See:

- https://www.ego4u.com/en/cram-up/writing/comma?10
- https://www.grammarly.com/blog/conditional-sentences/
- https://english.stackexchange.com/questions/6786/comma-in-conditional-sentence-and-in-antithesis

You can use `sed -n '/\<if\>/Ip' i18n/core/es.po` to get a list of all sentences that use the word 'if' or 'If'.

PR #1175, opened 2018-07-13
Mangopay exodus notification

Changaco:
WIP, feedback welcome.

### DIFF
--- a/emails/once/mangopay-exodus.spt
+++ b/emails/once/mangopay-exodus.spt
@@ -1,0 +1,29 @@
+{{ _("Liberapay is in trouble") }}
+
+[---] text/html
+<p>_("""\
+Our payment processor (Mangopay) is throwing us out, so we recommend that you {link1_start}evacuate the money from your account{link_end} before July 26th.
+
+Clicking on the link above will bring you to a new page titled “Emptying your wallet” and ask you to choose between two options.
+
+If you use Liberapay as a donor the first choice is recommended, especially if you wish to continue using Liberapay. This first option distributes all the money from your wallet to the people who would have normally received it over time. The recipients will be able to see that this sudden burst of income corresponds to multiple weeks of donations.
+
+For example, if you’re donating $1 per week to 2 creators, and you have $40 in your wallet, then each creator will receive $20, and 20 weeks from now you’ll receive our usual email reminding you that it’s time to add money again.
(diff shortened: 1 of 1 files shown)


LiamDawe:
Spelling error:

> recommanded

Should be recommended.

Other than that, it seems okay.

Changaco:
Thanks @LiamDawe.

Changaco:
I'm merging this now to give translators some time to do their work while I figure out how I'm going to proceed to actually send these notifications.

jorgesumle:
>  If you get an error message like “The attempt to distribute all the money in your wallet failed: XX.XX remains.” then the problem is probably on the recipient side 

It should read "If you get an error message like “The attempt to distribute all the money in your wallet failed: XX.XX remains.”**,** then the problem is probably on the recipient side"

> If you use Liberapay as a donor the first choice is recommended

It should read "If you use Liberapay as a donor**,** the first choice is recommended"

As of https://github.com/liberapay/liberapay.com/issues/959, I'd recommend not writing conditional sentences without comma. They are very difficult to read, incorrect and I always end up adding commas in the translation.

jorgesumle:
> You can find more information in {link3_start}this blog post{link_end}, and as always you can reply to this email if you need help.

The linked blog post from one of our strings is only in English, so translators should specify that it's only in English, but how can they know?

This is related to https://github.com/liberapay/salon/issues/136

Changaco:
I've just sent a first wave to 513 recipients.

```python
>>> from liberapay.main import *
Won't log to Sentry (SENTRY_DSN is empty).
>>> langs = set('nl fi es pt uk ru ms nb zh fr en'.split())
>>> recipients = website.db.all("select p from participants p where balance > 0 and is_suspended is not true and exists (select 1 from transfers t where t.tippee = p.id)")
>>> len(recipients)
593
>>> len([1 for p in recipients if p.email_lang not in langs])
80
>>> sent = 0
>>> for p in recipients:                                   
...   if p.email_lang in langs:
...      p.notify('once/mangopay-exodus', force_email=True, type='warning')
...      sent += 1
... 
>>> sent
513
```

Changaco:
An Italian translation has been added. German is still missing.

Changaco:
I've just sent the final wave: 357 emails.

PR #1179, opened 2018-07-18
Adapt payday

Changaco:
This branch attempts to make it possible to run payday today.


Changaco:
Payday is done for today, but more work is needed to evolve it to the next level. In particular the changes in this branch don't correctly handle team takes that were distributed in advance. In fact the current team system isn't really compatible with the "payment in advance" method.

PR #1181, opened 2018-07-20
Implement connecting a Stripe account

Changaco:
First step towards the support of Stripe as a payment processor.


Changaco:
I think Stripe's sign up form would confuse and discourage many potential users, it's not designed for a donation platform like Liberapay and it's bloated with JavaScript (it literally makes my Firefox lag).

jelv:
Also it asks you for your company name. What is a bit weird if you are not a company.

PR #1191, opened 2018-08-05
Stripe payments

Changaco:
This branch implements funding one-to-one donations through Stripe.

### DIFF
--- a/js/10-base.js
+++ b/js/10-base.js
@@ -38,6 +38,7 @@ Liberapay.init = function() {
     Liberapay.lookup.init();
     Liberapay.payments.init();
     Liberapay.s3_uploader_init();
+    Liberapay.stripe_init();
 
     $('div[href]').css('cursor', 'pointer').click(function() {
         location.href = this.getAttribute('href');
@@ -114,6 +115,17 @@ Liberapay.error = function(jqXHR, textStatus, errorThrown) {
     }
     Liberapay.notification(msg, 'error', -1);
 }
+
+Liberapay.wrap = function(f) {
+    return function() {
+        try {
+            return f.apply(this, arguments);
+        } catch (e) {
+            console.log(e);
+            Liberapay.notification(e, 'error', -1);
+        }
+    }
+};
 
 Liberapay.jsonml = function(jsonml) {
     var node  = document.createElement(jsonml[0]);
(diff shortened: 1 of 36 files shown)


Changaco:
Screenshots:

## For a new donation at the maximum amount

screenshot_20180806_135432

## For a new low amount donation, when the donor has already added a payment card

screenshot_20180806_134654

## When renewing multiple donations

screenshot_20180806_134642

LiamDawe:
Honestly, when doing the system like this, having it still state weekly when the payments are upfront just doesn't make sense.

Changaco:
Donors have always been asked to pay for many weeks upfront in order to minimize fees. The new Stripe payment forms visible in the screenshots above are quite similar to the old Mangopay forms.

LiamDawe:
I see, fair enough.

trebmuh:
Nice !

Changaco:
I'm having second thoughts about this implementation. We're liable for chargebacks but we don't have enough information to detect fraudulent payments. We can't even detect people who send money to themselves because we don't collect the name of the payer and Stripe doesn't give us access to the name of the recipient. At the very least we should verify the email addresses of both the payer and the recipient before we capture any payment.

LiamDawe:
Wow really? That's pretty basic stuff to be doing, thought you already had some basic checks in place.

Pretty sure most payment processors need you to perform some kind of checks.

Changaco:
With Mangopay we collected the name, birthday, nationality and country of residence of every user. Moreover identity verification was required for users who gave more than €2500 per year or received more than €1000 per year (legal thresholds).

With Stripe a significant amount of information is collected from the recipient when a Standard account is created, but we don't have access to this data, so we're in the dark and that's what I'm worried about. Moreover we're not collecting personal information from the payer because strictly speaking it's not required, but I'm worried that in this case collecting only what's required may not be enough.

Changaco:
This is deployed and it seems to be working. I've successfully made a payment and received a congratulatory email from Stripe for this first payment.

mattbk:
When I attempt to do this, I select the amount, enter card information, and click "Initiate the payment." Nothing seems to happen.

Also, an amount should be selected by default. I did not get an error when clicking the initiate button when no amount was selected.

Issue #1197, opened 2018-08-09
Receiving page information are confusing


trebmuh:
Related to: https://mastodon.xyz/@Liberapay/100520158923864989
Translation for the non-French readers:
> @Liberapay and otherwise, clarifying this mess would be possible???
> You need a stripe account to receive but not to give on one side, on another side it tells me that I'm still receiving????
> and from another side I can't give?
> and that, in fact, according to the logs, that... works?

***

On [my account](https://liberapay.com/me/receiving/), I've got this:

![liberapay-flou](https://user-images.githubusercontent.com/8705846/43909225-183b3448-9bfa-11e8-854d-77b25e8df6e2.png)

First line says that I'm receiving some money, second line says I can't receive money. It is confusing and needs to be clarified.

I'm trying a shot in the dark since I'm not totally sure how LP works nowadays (I'm still catching up  after 4 weeks of-the-web with what changed in LP):
- **First line**: You're receiving €XX.XX from donors per week.
- **Second line**: You can't transfer the money from your LiberaPay account to your bank account until you've configured a payment account.

@Changaco can you please confirm/infirm that my shot in the dark is right/wrong?

trebmuh:
( ping @liberapay/media-managers )

trebmuh:
(maybe) related to https://github.com/liberapay/liberapay.com/issues/1188

mattbk:
More accurately, it might be

- Donors want to give you $XX.XX per week.
- You can't receive money until you've connected an account to receive

(I think "payment account" in the context of receiving is confusing.)

No money is held in Liberapay; it goes straight from the donors' credit cards into the Stripe account.

trebmuh:
Thanks for the clarification @mattbk 
Trying #1198 to fix this, pending @Changaco approval.

PR #1198, opened 2018-08-09
Fix for #1197

trebmuh:
Fix for #1197

### DIFF
--- a/www/%username/receiving/index.html.spt
+++ b/www/%username/receiving/index.html.spt
@@ -24,15 +24,15 @@ teams = website.db.all("""
 
 % set receiving = participant.receiving
 % if participant == user
-<p>{{ _("You receive {0} per week.", receiving) }}</p>
+<p>{{ _("Donors want to give you {0} per week.", receiving) }}</p>
 % else
-<p>{{ _("{0} receives {1} per week.", participant.username, receiving) }}</p>
+<p>{{ _("Donors want to give {0} {1} per week.", participant.username, receiving) }}</p>
 % endif
 
 % if participant.is_person
 
 % if not participant.has_payment_account
-    <p>{{ _("You can't receive money until you've configured a payment account.") }}</p>
+    <p>{{ _("You can't receive money until you've connected an account to receive.") }}</p>
     <p><a class="btn btn-primary" href="{{ participant.path('payment') }}">{{ _("Configure payments") }}</a></p>
 % elif participant.balance > 0
     <p><a class="btn btn-default" href="{{ participant.path('wallet/payout/' + b64encode_s(request.path.raw)) }}"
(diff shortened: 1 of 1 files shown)


mattbk:
Don't feel like you have to use exactly my words from #1197.

trebmuh:
I'd say that if they make sense, then they are the right ones!

Changaco:
The `receiving` number being displayed isn't "how much donors want to give you", it's "how much you've received for the current week".

trebmuh:
I've tried a fix.
I'm not too sure if "received" is better than "are/is receiving" in this context.
@Changaco : please review.

Changaco:
I think that's still too confusing. The `receiving` number isn't just what the user received this week, it's also what they can expect to receive next week and every week after that. It's the user's stable income.

trebmuh:
@Changaco yes you're right, this is still confusing.

Let's make a example to clarify it.

User "bob" did have last week 10 donors who gave 0.5€ each. Then he earned 5€ on the **last  payday**. Let says here we are Monday. Since last Wednesday (= last payday), user "bob" got a new donor of 0.5€, then on next payday (=next Wednesday), he's expected to get 5.5€.

What is the `receiving` number to be displayed?

Changaco:
@trebmuh €5.50 (if I understood your question correctly).

trebmuh:
Thank you @Changaco .

So the sentence "You receive {0} per week." means "You are expecting to receive {0} for the next payday". For disambiguation, I propose that we should word it like that.

Then, I propose we rephrase the other sentence such as "But you can't receive money until you've connected an account to receive."

How does that sound to you @mattbk @Changaco and others?

www/%username/receiving/index.html.spt
```diff
@@ -24,15 +24,15 @@ teams = website.db.all("""
...
-<p>{{ _("{0} receives {1} per week.", participant.username, receiving) }}</p>
+<p>{{ _("{0} received {1} for the current week.", participant.username, receiving) }}</p>
 % endif
 
 % if participant.is_person
 
 % if not participant.has_payment_account
-    <p>{{ _("You can't receive money until you've configured a payment account.") }}</p>
+    <p>{{ _("You can't receive money until you've connected an account to receive.") }}</p>
     <p><a class="btn btn-primary" href="{{ participant.path('payment') }}">{{ _("Configure payments") }}</a></p>
```
mattbk: I think this should be phrased something like "Configure an account to receive payments".

mattbk:
@trebmuh, your first sentence may be even more confusing because weekly payday is going away--all Stripe and PayPal (+whatever) transactions happen at once.

If we're going to stick with the illusion of weekly support (and I disagree with that idea, see #1205), maybe "You've received donations that average $X for the next Y weeks"?

Changaco:
Closing in favor of #1297.

PR #1211, opened 2018-08-16
Update Debian version from Dockerfile

### LINKED ISSUE #1204: Error when installing dev env with docker
I have an error at the second step:

```
$ docker-compose run web bash recreate-schema.sh                                        ✔  2994  11:25:13
Creating network "liberapaycom_default" with the default driver
Creating liberapaycom_db_1 ... done
==============================================================================
Applying sql/recreate-schema.sql ...

psql: could not connect to server: Connection refused
	Is the server running on host "db" (172.24.0.2) and accepting
	TCP/IP connections on port 5432?
```

jorgesumle:
Now it uses Debian 9. @MartinDelille, this should fix #1204

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:8
+FROM debian:9
 
 RUN apt-get update && \
     apt-get install -y build-essential wget libpq-dev libffi-dev python-dev postgresql-client && \


MartinDelille:
Thanks @jorgesumle !

PR #1220, opened 2018-08-20
Add a signal user button

### LINKED ISSUE #353: No "report user" button
Having one would help reduce "noise" such as https://github.com/liberapay/salon/issues/45 and emails to support like the ones copied in https://github.com/liberapay/liberapay.org/issues/11, thus saving everyone time.

MartinDelille:
This is a work in progress. Feel free to comment the choice I made in the UI.

It aims to fix #353

### DIFF
--- a/style/base/base.scss
+++ b/style/base/base.scss
@@ -129,6 +129,16 @@ body {
     }
 }
 
+.btn-signal {
+    $base-color: #aaa;
+    background-color: $base-color;
+    color: #1a171b;
+    &:hover {
+        background-color: #f00;
+        color: #fff;
+    }
+}
+
 input.amount {
     min-width: 5em;
     max-width: 8em;
--- a/templates/profile.html
+++ b/templates/profile.html
@@ -97,6 +97,9 @@
                             {{ _("Donate") }}
                         </a>
                     % endif
+                    <a class="btn btn-signal btn-lg" href="{{ participant.path('signal') }}">
+                        {{ _("Signal") }}
+                    </a>
                 </div>
             </div>
         </div>
--- a/www/%username/signal.spt
+++ b/www/%username/signal.spt
@@ -1,0 +1,101 @@
+from liberapay.exceptions import UserDoesntAcceptTips
+from liberapay.utils import get_participant
+
+[---]
+participant = get_participant(state, restrict=False)
+if participant.goal == -1:
+    raise response.error(403, UserDoesntAcceptTips(participant.username))
+
+full_title = _("Signal {0}", participant.username)
+
+[---] text/html
+% extends "templates/profile-base.html"
+
+% from 'templates/elsewhere.html' import account_elsewhere with context
+
+% block head_early
+{{ super() }}
+    <meta property="og:description"
+          content="{{ _('Support {username}\'s work with a recurrent donation.',
+                        username=participant.username) }}">
+% endblock
+
+{% block heading %}{% endblock %}
+
+% block content
+<div class="row">
+<div class="col-md-10 col-md-offset-1 col-lg-8 col-lg-offset-2">
+    <h2>{{ _("Signal {0}", ('<a href="/{0}">{0}</a>'|safe).format(participant.username)) }}</h2>
+
+    % if not participant.hide_receiving
+        % set goal, receiving = participant.goal, participant.receiving
+        % if goal
+            % if receiving
+            <p>{{ _(
+                "{0} currently receives {1} per week, they need your help to reach "
+                "their funding goal ({2} per week)."
+                , participant.username, receiving, goal
+            ) }}</p>
+            % else
+            <p>{{ _(
+                "{0}'s goal is to receive {1} per week."
+                , participant.username, goal
+            ) }} {{ _("Be the first to contribute!") }}</p>
+            % endif
+        % else
+            <p>{{ _(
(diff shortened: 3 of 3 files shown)


mattbk:
Do you mean "report" a user?

Need to add documentation for users as well.

LiamDawe:
The wording is "report" not signal as @mattbk mentioned, signal doesn't really make sense as people know what reporting is.

Also, spelling/grammar issues:

> What happen when I signal a user

Should be "What happens when I report a user?"

> The profil will be review by the moderator. 

Should be "The profile will be reviewed by a moderator."

mattbk:
I think a link to more detail is needed. Who are the moderators, what powers do they have, what's the process, etc.

MartinDelille:
@mattbk I updated the FAQ accordingly to the proposition made there: https://github.com/liberapay/liberapay.org/issues/15#issuecomment-414093827

www/%username/report.spt
```diff
@@ -96,6 +96,19 @@ full_title = _("Report {0}", participant.username)
         "The profil will be reviewed by a moderator. "
     ) }}</p>
 
+    <h4>{{ _("Who are the moderators?") }}</h4>
+    <p>{{ _(
+        "There is ten moderators who have been elected by the ten most funded project on Liberapay. "
```
mattbk: *platform's

www/%username/report.spt
```diff
@@ -0,0 +1,114 @@
...
+
+    <h4>{{ _("Who are the moderators?") }}</h4>
+    <p>{{ _(
+        "There is ten moderators who have been elected by the ten most funded project on Liberapay. "
+        "This ensure that Liberapay moderation is made by independent people concerned by the plateform safety. "
+    ) }}</p>
+
+    <h4>{{ _("What is the moderating process?") }}</h4>
+    <p>{{ _(
+        "Once at least half of the moderators confirm that the user is suspicious,"
```
mattbk: Account, not user, if you use "its" in the next line.

www/%username/report.spt
```diff
@@ -0,0 +1,114 @@
...
+        "its account is closed and it can not received donation anymore. "
```
mattbk: *receive

www/%username/report.spt
```diff
@@ -0,0 +1,114 @@
...
+        "All donations to the user are then cancelled."
```
mattbk: Or account, see above.

www/%username/report.spt
```diff
@@ -0,0 +1,114 @@
...
+        ) }}</p>
+        % for platform in website.platforms if platform.name in e_accounts
+            {{ account_elsewhere(platform, e_accounts, None) }}
+        % endfor
+        <br>
+    % endif
+
+    % block form
+    <p>{{ _(
+        "Please add a short description why you want to report this user:"
```
mattbk: "Please describe why..."

www/%username/report.spt
```diff
@@ -0,0 +1,114 @@
...
+        </div>
+
+        <button class="report btn btn-success" name="report" value="true">{{ _("Report") }}</button>
+    % endblock
+
+    <h3>{{ _("Frequently Asked Questions") }}</h3>
+
+    <h4>{{ _("What happens when I report a user?") }}</h4>
+    <p>{{ _(
+        "The profil will be reviewed by a moderator. "
```
mattbk: *profile

www/%username/report.spt
MartinDelille: What do you mean?

MartinDelille:
How can I go further with this?

I would say we should add a kind of *role* attribute to a user to differenciate a regular user from a moderator.

Then a simple email or notification would be sent to the 10 moderators that would be triggered everytime a user click the report button.

MartinDelille:
@Changaco Any thought about this?

mattbk:
Away from computer right now, but admins in Gratipay before this was forked were distinguished by a column in the participant table called is_admin. Then testing whether a user was an admin was possible by querying the current user.

A similar system could be set up for mods if this is still true in Liberapay.

mattbk:
Then you would need a voting mechanism for reports and to record who voted what or a simple count.

Additional feature would be discussion of reports on top of voting. 

But for a quicker release, quarantining a suspicious user could be done manually by an admin.

MartinDelille:
Your right: just notifying the admin would be a minimum.

I recently report a facebook page and they have a pretty way to handle it: When the moderator have examined the suspicious page, a report is sent to the user (but like you say, this would be additional features).

Changaco:
I replaced the `is_admin` column with a more generic `privileges` column 2 years ago in 63900b45385bfb62be105c324cc53d45154e3c57. I have a local branch to add a new `moderate` privilege, but it's not ready yet.

MartinDelille:
@Changaco Awesome! I guess the unknown part I have to do when the button clicked is something like:

```python
warning = "The following account has been reported:" + reportedAccount

foreach user in users where user.privileges & moderate != 0
    notifications.sent(user, warning)
```

Any code example I could go from to:
- do the moderator user request
- send the notification

mattbk:
Would anything from liberapay.com/tests/py/test_notifications.py help you out?

Should probably add a test there anyway.

www/%username/report.spt
Cher172: Why??

PR #1238, opened 2018-09-13
2 typos

trebmuh:
Found with `codespell -q 3`

### DIFF
--- a/js/mangopay.js
+++ b/js/mangopay.js
@@ -198,7 +198,7 @@ var mangoPay = {
                 // Invoke the user supplied success or error handler here
                 success: function(data, xmlhttp) {
 
-                    // Parse API reponse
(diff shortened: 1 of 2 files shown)


Changaco:
I don't think it's a good idea to fix typos in those files, especially since they're going to be deleted in the future anyway.

trebmuh:
> I don't think it's a good idea to fix typos in those files

Can I ask why? It's unclear for me.

Changaco:
Like I said these files are going to be deleted anyway. Also they're not really Liberapay code, they're third party files included in the repo, so they should only be modified if it's really necessary and the modifications aren't going to cause more problems later.

trebmuh:
That makes sense. Thank you.
Closing then.

PR #1267, opened 2018-10-07
Improve the "Payment" page

Changaco:
This branch renames the "Payment" page to "Payment Processors" and adds more information to it.


trebmuh:
Checked quickly what will be showed to the user, it looks fine and understandable so far.

PR #1272, opened 2018-10-10
Tweak the income notification wording

Changaco:
Is it any clearer?

### DIFF
--- a/emails/income~v2.spt
+++ b/emails/income~v2.spt
@@ -29,8 +29,8 @@
 
 <p>{{ _(
     "NB: Donations through Liberapay are now paid in advance instead of being transferred "
-    "weekly. The numbers above match what you would have received this week under the old "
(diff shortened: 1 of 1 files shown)


trebmuh:
Can you explain the difference between "for" and "during" (e.g. within an asterisk)?

Changaco:
Well, that answers my question: it's not clearer.

trebmuh:
Correct. The hard part here I guess is to make the information clear from an user point of view. Probably the information you want to write here is clear in your brain because you wrote the code. But it's still not from an user POV. Probably a couple of sentences will be needed here to make it crystal clear.

Changaco:
The paragraph I was trying to fix in my first commit isn't the real problem, it's the one below mentioning the Mangopay wallet balance that generates the most confusion.

Issue #1323, opened 2018-11-07
Raising amount possible bug?


trebmuh:
My actual take on the LP team is 5€ a week.

The following is displayed on liberapay.com when I'm logged in:

![bug2-lp](https://user-images.githubusercontent.com/8705846/48136092-14b6c580-e29f-11e8-8ada-9cf29c3f0683.png)

According to https://liberapay.com/about/teams#throttling , I shouldn't be able to ask for more than a 10€ take for next week if the "Regulation of take amounts" functionality is activated on the LP team. If this functionality isn't activated on the LP team, then I don't know where the _26,50€_ amount calculation comes from since (at the time of writing), the LP team leftover is _5,00 € and 2,93 $US_ which would make the following calculation for the maximum amount I could ask for: 5€ (week take) + 5€ + 2.93$ which equals to roughly a bit less than half of the _26,50€_ mentioned above.

It doesn't looks clear to me what's going on, and I might be totally/partially wrong with how the system works, but something sounds fishy here.

Changaco:
€26.50 is indeed a weird limit given the current situation. Looking at the code I see a bug that might be responsible.

Changaco:
After fixing that bug I'm getting €21.65, which is €5.00 + a leftover of €16.65.

trebmuh:
Does that corresponds to the actual leftover of the team? _(I guess you're trying that on a home-made installation of LP which might not be on the same state than the production one)_

trebmuh:
OK

trebmuh:
Going further, how does a team can enable the take throttling?

Changaco:
Adding or removing take throttling requires sending an email to support. I remember this happening at least once.

Changaco:
The remaining TODO before closing this issue is to figure out why the team's estimated income is fluctuating between ~€160 and ~€190, thus resulting in a different leftover and different take limits.

trebmuh:
> Adding or removing take throttling requires sending an email to support. I remember this happening at least once.

I've made a PR to add this knowledge to the https://liberapay.com/about/teams page. See #1330 .

trebmuh:
(up)

Looking at https://liberapay.com/me/receiving/ , I'm receiving 10€/week, there is ~61,59 € leftover for the team, and the page tells me that I can take up to €119.99 this week:

![lp-upto](https://user-images.githubusercontent.com/8705846/103940614-e73f5e00-512d-11eb-9e1e-9950a0b03833.png)

[...]

PR #1330, opened 2018-11-09
Adding info about "team throttling" in about/teams

trebmuh:
See #1323

### DIFF
--- a/www/about/teams.spt
+++ b/www/about/teams.spt
@@ -130,8 +130,9 @@ title = _("Teams")
 ) }}</p>
 
 <p>{{ _(
-    "You can enable this option when you create a team. "
-    "If you want to change this option later, you need to contact our support."
+    "Please contact support if you want to enable or disable the take limits "
+    "for an existing team. Don't forget to include evidence that the change "
+    "has been approved by the other team members."
 ) }}</p>
 
 <h3>{{ _("Removing team membership") }}</h3>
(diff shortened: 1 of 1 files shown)


www/about/teams.spt
```diff
@@ -129,6 +129,11 @@ title = _("Teams")
     "been created after the most recent payday)."
 ) }}</p>
 
+<p>{{ _(
+    "You can enable or not this option a the very moment you're creating a team. "
+    "If you want to change this option later, you need to contact our support."
```
mattbk: Include a link to email?

www/about/teams.spt
```diff
@@ -129,6 +129,11 @@ title = _("Teams")
     "been created after the most recent payday)."
 ) }}</p>
 
+<p>{{ _(
+    "You can enable or not this option a the very moment you're creating a team. "
```
mattbk:
```suggestion
    "You can enable this option when you create a team. "
```

www/about/teams.spt
trebmuh: I thought about that when writing it but I saw there is another "contact" in the "Removing team membership" of the https://liberapay.com/about/teams page, plus there is a "contact" tab at the beginning of the same page, so I ended not putting a link.

trebmuh:
Thanks for the review and comment @mattbk .
@Changaco how does that looks for you?

www/about/teams.spt
mattbk: Right on.

Changaco:
I've reworded the paragraph to be more consistent with the rest of the page, please review.

trebmuh:
LGTM

trebmuh:
:+1:

PR #1348, opened 2018-11-19
Code profiling and optimization

Changaco:
This branch adds profiling tools and optimizes a few hot spots.


Changaco:
Profiling graph of rendering the homepage a thousand times

Changaco:
Profiling graph of serving a small static file 10000 times

PR #1372, opened 2019-01-04
Fraud review

Changaco:
This branch brings back the UI to block payments to an account that looks fraudulent, and creates a new admin tool to facilitate fraud review.

### DIFF
--- a/liberapay/i18n/base.py
+++ b/liberapay/i18n/base.py
@@ -9,7 +9,7 @@ from unicodedata import combining, normalize
 from six import text_type
 
 import babel.core
-from babel.dates import format_date, format_datetime, format_timedelta
+from babel.dates import format_date, format_datetime, format_time, format_timedelta
 from babel.messages.pofile import Catalog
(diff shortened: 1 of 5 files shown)


Changaco:
Screenshot of the new admin tool for fraud review

Changaco:
Background: when Liberapay launched 3 years ago its anti-fraud mechanism was the quarantine of funds for one month, but that required holding funds in escrow and didn't fully stop fraud, so now we're moving towards a systematic review of all payment recipients.

Since Liberapay is still "neutral" we will only block payments when the user appears to be involved in fraud or another illegal activity, whereas Gratipay 2.0 had other criteria like "brand fit" when reviewing new projects.

mattbk:
>  Gratipay 2.0 had other criteria like "brand fit" when reviewing new projects.

That was a separate process from the fraud review.

Changaco:
I know, but the public review also resulted in payments being blocked if the project was rejected.

The point was that Liberapay will still be different from Gratipay 2.0.

PR #1373, opened 2019-01-04
Log admin actions

Changaco:
When an admin modifies a user's account, this action needs to be logged, both for accountability and to avoid confusion on who did what.


Changaco:
Screenshot of an admin page showing a logged admin request

PR #1387, opened 2019-01-16
Adapt payday stats to Liberapay 2.0

### LINKED ISSUE #1327: The "Stats" page needs to be adapted
On [the stats page](https://en.liberapay.com/about/stats), it's written "_€23,253.40 and $2,630.25 is escrowed within Liberapay._" which can give the impression that those >23k€ and >2k$ are stocked/blocked **in** liberapay.

I'm not too sure about the reality of this. It can well be the case if that amount of money is what's left in the different mangopay users wallets but it seems to me like a big amount. Then, I'm thinking that it might probably be a figure of the money which have been given by donors to receivers for the next weeks/months to come.

In the case I'm right with the second hypothesis, I'd suggest another wording such as "_In Liberapay, donors are actually giving 23 253,40 € and 2 630,25 $US on the weeks to come._"

_(BTW, there is no mention of the few PLN which have starting to come on last payday, yesterday)_

Changaco:
Closes #1327. `https://liberapay.com/about/stats-v1` will allow seeing the old stats page for a while.

### DIFF
--- a/liberapay/billing/payday.py
+++ b/liberapay/billing/payday.py
@@ -813,6 +813,13 @@ class Payday(object):
                         WHERE amount < 0
                           AND refunded
                    )
+                 , week_payins = (
+                       SELECT basket_sum(pi.amount)
+                         FROM payins pi
+                        WHERE pi.ctime < %(ts_start)s
+                          AND pi.ctime >= %(previous_ts_start)s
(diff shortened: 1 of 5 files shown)


Changaco:
Screenshot of the new charts

PR #1393, opened 2019-01-22
Encrypted storage of identity information

Changaco:
This branch adds the infrastructure needed to store sensitive information in our database. It's inspired by Gratipay's implementation of encrypted storage.


Changaco:
The next step is to figure out what information we're going to store. With Mangopay there was no choice, we had to collect the info they required, but Stripe and PayPal collect the info they want through their own websites.

On one hand I'd like to collect as much information as possible for anti-fraud purposes, but on the other hand storing data we don't really need could be considered a violation of the GDPR.

PR #1395, opened 2019-01-25
Prepare to payout remaining balances indirectly

Changaco:
The idea is to transfer the funds into the LiberapayOrg wallet, then withdraw them to Liberapay's bank account, and finally pull them into PayPal for a MassPay.


tflex187:
$$$$

PR #1431, opened 2019-02-05
Pledge notifications

### LINKED ISSUE #795: Pledge activation notifications can be misleading
1. The message doesn't link to the actual account the donation has been attributed to, which can lead to confusion.
2. Donors aren't told if they need to add money to fund their new donations.

Changaco:
This branch fixes #795.


Changaco:
I went a little too fast with this: a test on the production DB reveals that the new SQL query introduced by this PR returns false positives.

PR #1432, opened 2019-02-06
Document alternatives

Changaco:
This branch is for #269.

So far I only laid down a basic structure without including any real content. Feel free to contribute.

### DIFF
--- a/templates/nav.html
+++ b/templates/nav.html
@@ -36,6 +36,7 @@
         ('/global', _('Global')),
         ('/teams', _('Teams')),
         ('/stats', _('Stats')),
+        ('/alternatives', _('Alternatives')),
         ('/legal', _('Legal')),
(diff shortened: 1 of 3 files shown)


trebmuh:
I've added Tipeee with PR #1437 .
Wondering if we should add gratipay in as well?

mattbk:
Gratipay no longer exists.

trebmuh:
Yes I know that.
A lot of people still know that Liberapay takes his origin from it, and then could be easily eager for information between both.

Changaco:
I've added Ghost because of [this tweet](https://twitter.com/TeddyFERDINAND1/status/1473736724337643522) and because [Ghost compares itself to Patreon](https://ghost.org/vs/patreon/) (and [a bunch of other platforms](https://ghost.org/alternatives/)).

PR #1439, opened 2019-02-10
Make 4xx errors more user-friendly

### LINKED ISSUE #290: 4xx error pages aren't user-friendly
We should add a message on 4xx pages explaining to the user that it looks like the error came from their side, and that if they believe the problem is actually on our side they should open an issue here on github or send us an email.

kindlyfire:
Improvements to the messages shown for 4xx errors. Trying to fix #290

### DIFF
--- a/error.spt
+++ b/error.spt
@@ -28,12 +28,32 @@ if code == 500 and not err:
     % if response.html_template
         % include response.html_template
     % else
-        <p>{{ msg }}</p>
-        % if website.show_tracebacks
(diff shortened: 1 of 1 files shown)


Changaco:
I've reworded the new messages and added some debugging info.

mattbk:
Thanks @kindlyfire!

trebmuh:
Agreed, thanks @kindlyfire and @Changaco !

PR #1440, opened 2019-02-10
Update MangoPay reference

### LINKED ISSUE #1401: Remove MangoPay References
E.g. here (https://liberapay.com/about/legal). From my understanding they are no longer working with Liberapay.

trebmuh:
close #1401


Changaco:
Now that the paragraph has been properly reworded should we show it to everyone or keep it hidden by default?

trebmuh:
What about showing it only to users with a MangoPay account? Is that possible?

PR #1449, opened 2019-02-14
Fix notifications

Changaco:
I messed up in #1431. Last night the new `notify_patrons` cron started sending the same notification over and over again to 3 users. Each of them received 148 identical emails!

This branch fixes the bug and adds a check to prevent similar bugs in the future.


trebmuh:
Ouch! :smiley: 
Thank for acting quickly!

PR #1455, opened 2019-02-23
Drop pickle

### LINKED ISSUE #1132: Usage of `pickle` is potentially dangerous
We use python's `pickle` module to store notification contexts as bytes in the database. If an attacker was somehow able to inject an arbitrary value then that would lead to a Remote Code Execution vulnerability.

Report (not yet public): <https://hackerone.com/reports/361341>.

### LINKED ISSUE #1414: Pickle fails to load old notification contexts
https://sentry.io/share/issue/e6bc3189f12d4be9aca8e9ecce9eb189/

```
  File "liberapay/models/participant.py", line 1266, in render_notifications
    notif_context = deserialize(notif_context)
  File "liberapay/utils/__init__.py", line 387, in deserialize
    return pickle.loads(context)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 1: ordinal not in range(128)
```

Changaco:
Second and final step of the migration of notification data from pickle to CBOR (previous: #1454). Closes #1132, fixes #1414.

### DIFF
--- a/liberapay/i18n/currencies.py
+++ b/liberapay/i18n/currencies.py
@@ -223,19 +223,6 @@ class MoneyBasket(object):
 
     __nonzero__ = __bool__
 
-    def __setstate__(self, state):
-        """Backward-compatible unpickling
-
-        The original version of `MoneyBasket` stored `Money` objects in its
-        `__dict__`, whereas the current version stores `Decimal`s in the
-        `amounts` attribute.
-        """
-        if 'amounts' in state:
-            self.amounts = state.pop('amounts')
-            self.__dict__ = state
-        else:
-            self.amounts = {m.currency: m.amount for m in state.values()}
-
     @property
     def currencies_present(self):
         return [k for k, v in self.amounts.items() if v > 0]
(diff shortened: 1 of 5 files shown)


Changaco:
Actually this is the third step, the second was the actual data migration that I performed manually.

Here's a summary of how I did the migration:

1. I downloaded the notifications that needed to be migrated, into a CSV file.
2. I created a second clone of the liberapay.com repo on my machine, checked out commit ded842dfe46071df5be86f993bfa062d6ca1fec7 to get a version that would run under Python 2.7, and backported some changes from #1454.
3. I opened a Python shell and threw a bunch of commands at it to turn the CSV file I had downloaded into an SQL file containing lots of `UPDATE notifications SET context = ..., context_is_cbor = true WHERE id = ...;` lines.
4. I applied the changes to the production DB.
5. At this point pretty much all notifications older than 2019-01-28 were broken, because I forgot to monkey-patch the CBOR encoder to treat Python 2's strings as text instead of bytes.
6. I redid step 3 and 4 with the patched CBOR encoder.
7. I did something similar to step 3 in production for all the notifications created on or after 2019-01-28 (the day we switched to Python 3).

PR #1470, opened 2019-03-16
Protect community from confusable homoglyphs

### LINKED ISSUE #1469: Confusable homoglyphs
When we allow Unicode in names, like we do for community names, we should reject names that look like other names: https://github.com/vhf/confusable_homoglyphs

MartinDelille:
This PR aims to fix #1469 by protecting the creation of confusable name for community.

I learned how to use the library during the implementation but there is something I don't know very well how to do.

The documentation says that we have to [update the data file](https://confusable-homoglyphs.readthedocs.io/en/latest/usage.html) by executing the two following commands:

```
pip install confusable_homoglyphs[cli]
confusable_homoglyphs update
```

I suppose the call the `confusable_homoglyphs update` has to be added to the `Makefile` during the `make env` step, am I right?

### DIFF
--- a/liberapay/models/community.py
+++ b/liberapay/models/community.py
@@ -3,6 +3,7 @@ import unicodedata
 
 from postgres.orm import Model
 from psycopg2 import IntegrityError
+from confusable_homoglyphs import confusables
 
 from liberapay.exceptions import CommunityAlreadyExists, InvalidCommunityName
 
@@ -38,8 +39,17 @@ class Community(Model):
         name = unicodedata.normalize('NFKC', name)
         if name_re.match(name) is None:
             raise InvalidCommunityName(name)
+
         try:
             with cls.db.get_cursor() as cursor:
+                all_names = cursor.all("""
+                    SELECT name
+                    FROM communities
+                    """)
+                for existing_name in all_names:
+                    if cls._unconfusable(name) == cls._unconfusable(existing_name):
+                        raise CommunityAlreadyExists
+
                 p_id = cursor.one("""
                     INSERT INTO participants
                                 (kind, status, join_time)
@@ -91,3 +101,14 @@ class Community(Model):
     @property
     def nsubscribers(self):
         return self.participant.nsubscribers
+
+    @staticmethod
+    def _unconfusable(name):
+        unconfusable_name = ''
+        for c in name:
+            confusable = confusables.is_confusable(c, preferred_aliases=['COMMON', 'LATIN'])
+            if confusable:
+                # if the character is confusable we replace it with the first prefered alias
+                c = confusable[0]['homoglyphs'][0]['c']
+            unconfusable_name += c
+        return unconfusable_name
--- a/tests/py/test_communities.py
+++ b/tests/py/test_communities.py
@@ -1,6 +1,9 @@
 import json
 
-from liberapay.exceptions import AuthRequired
+from liberapay.exceptions import (
+        AuthRequired,
+        CommunityAlreadyExists,
+)
 from liberapay.models.community import Community
 from liberapay.testing import Harness
 
@@ -100,6 +103,30 @@ class TestCommunityActions(Harness):
 
         communities = self.bob.get_communities()
         assert len(communities) == 0
+
+    def test_create_community_already_taken(self):
+        with self.assertRaises(CommunityAlreadyExists):
+            Community.create('test', self.alice.id)
+
+    def test_create_community_already_taken_is_case_insensitive(self):
+        with self.assertRaises(CommunityAlreadyExists):
+            Community.create('TeSt', self.alice.id)
+
+    def test_unconfusable(self):
+        self.assertEqual('user2', Community._unconfusable('user2'))
+        self.assertEqual('alice', Community._unconfusable('alice'))
+        latin_string = 'AlaskaJazz'
+        mixed_string = 'ΑlaskaJazz'
+        self.assertNotEqual(latin_string, mixed_string)
+        self.assertEqual(latin_string, Community._unconfusable(mixed_string))
+
+    def test_create_community_already_taken_with_confusable_homoglyphs(self):
+        latin_string = 'AlaskaJazz'
+        mixed_string = 'ΑlaskaJazz'
+
+        Community.create(latin_string, self.bob.id)
+        with self.assertRaises(CommunityAlreadyExists):
+            Community.create(mixed_string, self.alice.id)
 
 
 class TestCommunityEdit(Harness):


mattbk:
How often do they need to be updated? Could this be done in cron, or in the payday script?

MartinDelille:
I think updating it every time the site is updated would be fair enough. The script update if from this page: http://ftp.unicode.org/Public/security/latest/confusables.txt

Last update was november 2018.

liberapay/models/community.py
```diff
@@ -3,6 +3,7 @@
 
 from postgres.orm import Model
 from psycopg2 import IntegrityError
+from confusable_homoglyphs import confusables
```
Changaco: Please keep imports sorted alphabetically (within each group).

liberapay/models/community.py
```diff
@@ -38,8 +39,17 @@ def create(cls, name, creator_id, lang='mul'):
...
         try:
             with cls.db.get_cursor() as cursor:
+                all_names = cursor.all("""
+                    SELECT name
+                    FROM communities
+                    """)
+                for existing_name in all_names:
+                    if cls._unconfusable(name) == cls._unconfusable(existing_name):
+                        raise CommunityAlreadyExists
+
```
Changaco: This is a very inefficient implementation, the more communities there are the slower it becomes. Also you're repeating an identical function call (`cls._unconfusable(name)`) in every loop iteration.

liberapay/models/community.py
```diff
@@ -91,3 +101,14 @@ def check_membership_status(self, participant):
...
+    @staticmethod
+    def _unconfusable(name):
+        unconfusable_name = ''
+        for c in name:
+            confusable = confusables.is_confusable(c, preferred_aliases=['COMMON', 'LATIN'])
+            if confusable:
+                # if the character is confusable we replace it with the first prefered alias
+                c = confusable[0]['homoglyphs'][0]['c']
+            unconfusable_name += c
+        return unconfusable_name
```
Changaco: This function isn't specific to communities, so it should probably be in the `liberapay.utils` module instead of in the `Community` class.

requirements_base.txt
```diff
@@ -243,3 +243,6 @@ cryptography==2.4.2 \
     asn1crypto==0.24.0 \
         --hash=sha256:2f1adbb7546ed199e3c90ef23ec95c5cf3585bac7d11fb7eb562a3fe89c64e87 \
         --hash=sha256:9d5c20441baf0cb60a4ac34cc447c6c189024b6b4c6cd7877034f4965c464e49
+confusable_homoglyphs==3.2.0 \
+    --hash=sha256:3b4a0d9fa510669498820c91a0bfc0c327568cecec90648cf3819d4a6fc6a751 \
+    --hash=sha256:e3ce611028d882b74a5faa69e3cbb5bd4dcd9f69936da6e73d33eda42c917944
```
Changaco: You forgot to add a blank line above this new dependency block.

tests/py/test_communities.py
```diff
@@ -1,6 +1,9 @@
 import json
 
-from liberapay.exceptions import AuthRequired
+from liberapay.exceptions import (
+    AuthRequired,
+    CommunityAlreadyExists,
+)
```
Changaco: There's no need to split this import into multiple lines, `from liberapay.exceptions import AuthRequired, CommunityAlreadyExists` is less than 80 chars long.

Changaco:
> I'm not certain `make env` gets run very often in production unless needed by code changes.

`make` is not used at all in production.

Changaco:
We should probably have a copy of the `confusables.txt` file in the repo so that we can fall back to it if fetching the current file from the ftp.unicode.org server fails.

Changaco:
Strike that, the `confusable_homoglyphs` package already ships with a copy of the data transformed into JSON, so we don't need to deal with fetching the `confusables.txt` file at all.

liberapay/models/community.py
MartinDelille: delete here: 9b496bc88f8a3c404394ee8c1999e33a6f944a0b

liberapay/models/community.py
MartinDelille: Fixed here: 9b496bc88f8a3c404394ee8c1999e33a6f944a0b

requirements_base.txt
MartinDelille: Fixed here: 03b1fa39f628449ed1ccb2baeea4841d7c6fb231

tests/py/test_communities.py
MartinDelille: Fixed here: 2a96f25164ea687f67ee6a9915a485e9d3950f7b

MartinDelille:
Ok I removed the update in the `make env` step here: 64033bddaf8e7e8ef071831dd9c976379f1b46f2

liberapay/models/community.py
MartinDelille: Fixed here: 9b496bc88f8a3c404394ee8c1999e33a6f944a0b

liberapay/utils/unconfusable.py
```diff
@@ -0,0 +1,11 @@
+from confusable_homoglyphs import confusables
+
+def unconfusable_string(name):
```
Changaco: It would be nice to have a docstring explaining what this function does.

liberapay/models/community.py
MartinDelille: Which part is the most inefficient:

- the database query?
- the unconfusable lookup?
- the `unconfusable_string` function?

liberapay/utils/unconfusable.py
MartinDelille: Done here: a73af92

Tell me if there is a specific format for that.

liberapay/models/community.py
Changaco: The fetching of all the existing community names from the database and the subsequent loop. That's linear complexity, [O(n)](https://en.wikipedia.org/wiki/Big_O_notation), with `n` being the number of existing communities.

liberapay/utils/unconfusable.py
```diff
@@ -0,0 +1,13 @@
+from confusable_homoglyphs import confusables
+
+# Convert an Unicode string to its equivalent replacing all confusable homoglyphs
+# to its common/latin equivalent
```
Changaco: That's a comment, not a [docstring](https://en.wikipedia.org/wiki/Docstring).

liberapay/models/community.py
Changaco: We should probably do something like this instead:

```python
                possible_collisions = get_confusables(name)
                possible_collisions.append(name)
                collides_with = cursor.one("""
                    SELECT name
                      FROM communities
                     WHERE name IN %s
                     LIMIT 1
                """, (tuple(possible_collisions),))
                if collides_with:
                    raise CommunityAlreadyExists(collides_with)
```

PR #1509, opened 2019-05-24
Create a security.txt file.

EdOverflow:
This commit adds a security.txt file under the `.well-known` directory of liberapay.com. The file consists of a `Contact:` and `Acknowledgments:` directive that point to Liberapay's HackerOne program.

### DIFF
--- a/www/.well-known/security.txt
+++ b/www/.well-known/security.txt
@@ -1,0 +1,2 @@
+Contact: https://hackerone.com/liberapay
+Acknowledgments: https://hackerone.com/liberapay/thanks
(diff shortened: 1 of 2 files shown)


Changaco:
This doesn't work yet. The version of Aspen we're using still rejects all requests for "hidden" paths, even `.well-known`.

EdOverflow:
I had a feeling that would be the case, @Changaco. Any pointers on where in Aspen I could resolve this?

Changaco:
The only clean way to resolve this is to modify Pando to use the latest version of Aspen (1.0rc4). I have a branch for that and I'm looking to see if I can complete it today.

mattbk:
> This doesn't work yet.

You know, I had a feeling when I reviewed that I would be proven wrong. I need to trust that more.

PR #1518, opened 2019-06-11
Update the list of countries supported by Stripe

Changaco:
The "new" countries are currently still in a preliminary phase, you have to request an invite, so I'm not sure if we should consider them supported or not.


Changaco:
This would be confusing because the Stripe account creation form doesn't support the "new" countries. We should add a simple link to https://stripe.com/global#signup-form instead.

PR #1528, opened 2019-07-10
Update Japanese translation

hakaba-hitoyo:
(Title only)


Changaco:
Translations should only be submitted through Weblate, not in pull requests.

PR #1530, opened 2019-07-15
Email disavowal

### LINKED ISSUE #61: Prevent verification email spam
Report: https://hackerone.com/reports/87531

Simply copying [Gratipay's solution](https://github.com/gratipay/gratipay.com/commit/b4dc7eff5d141c9949aa959fe2598dab22621d76) won't be enough, because it doesn't entirely fix the problem: you can avoid the throttling by deleting and re-adding the address. (cc @whit537)

Also, I don't think adding a fixed delay is a good enough solution. At the very least the user being spammed needs an immediate way to stop the spam (a "I didn't request this, stop sending me emails" button). Stricter throttling would also be nice.

Changaco:
This branch closes #61.


Changaco:
Screenshot of the new email disavowal page

PR #1543, opened 2019-08-04
Avatar check

Changaco:
This branch drops the usage of automatically-generated avatars and adds a basic check of avatar URLs, so that we'll be able to warn users who don't have an avatar that their profiles are incomplete.


Changaco:
Now that this branch has been deployed I've launched a check of 25411 existing avatar URLs.

PR #1549, opened 2019-08-14
Refunds

Changaco:
This branch finally implements recording refunds and handling disputes automatically for Stripe payments. In the old wallet system refunds were recorded as negative payments, but this branch implements a different approach which is close to Stripe's, hopefully it's the right choice.

### DIFF
--- a/emails/payin_refund_initiated.spt
+++ b/emails/payin_refund_initiated.spt
@@ -1,0 +1,22 @@
+{{ _("A refund of {money_amount} has been initiated", money_amount=refund_amount) }}
+
+[---] text/html
+<p>{{ _(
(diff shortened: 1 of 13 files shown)


Changaco:
This is deployed, and I've run the following code to record all existing refunds:

```python
import stripe

from liberapay.payin.stripe import settle_charge

refunds = stripe.Refund.list()
for refund in refunds.auto_paging_iter():
    charge = stripe.Charge.retrieve(refund.charge)
    payin = website.db.one("SELECT * FROM payins WHERE id = %s", (charge.metadata['payin_id'],))
    print(refund.id)
    print(settle_charge(website.db, payin, charge))
```

PR #1575, opened 2019-09-10
Update Weblate logo

nijel:
The branding has changed some time ago, see
https://weblate.org/news/archive/weblate-changing-its-look/


Changaco:
@nijel Would you mind replacing the image with a version exactly 40px high? That's the size the image is displayed at. I looked for a scalable or high-res source to generate the 40px image myself but couldn't find one.

nijel:
No problem, I've just updated it (I didn't really looked how it's used, I tried to produce similar size as previous one).

In case you prefer to use SVG, it's available here: https://github.com/WeblateOrg/graphics/tree/master/logo-text

Changaco:
Perfect, thanks.

PR #1580, opened 2019-09-17
Pando 0.47

Changaco:
This branch upgrades to the new versions of Pando and Aspen.


Changaco:
With these updates and the other changes in this branch the tests run approximately 10% faster on my machine.

PR #1606, opened 2019-10-30
Add India to the list of countries supported by Stripe

Changaco:
It looks like Stripe is now available in India, however according to https://stripe.com/global it's still a "Preview". It would be nice to have confirmation, either from Stripe or from someone in India, that Stripe is indeed operational in that country, before merging this.

### DIFF
--- a/liberapay/constants.py
+++ b/liberapay/constants.py
@@ -351,8 +351,8 @@ PAYOUT_COUNTRIES = {
     """.split()),  # https://www.paypal.com/us/webapps/mpp/country-worldwide
 
     'stripe': set("""
-        AT AU BE CA CH DE DK EE ES FI FR GB GR HK IE IT JP LT LU LV MX MY NL NO
-        NZ PL PT SE SG SI SK US
+        AT AU BE CA CH DE DK EE ES FI FR GB GR HK IE IN IT JP LT LU LV MX MY NL
+        NO NZ PL PT SE SG SI SK US
         PR
     """.split()),  # https://stripe.com/global
 }


trinitronx (dismissed):
Looks good, simple enough 😄

Changaco:
There are currently 7 Stripe India accounts connected to Liberapay accounts, however the only payment attempt so far has failed with the error message “This account isn't enabled to make cross border transactions. You can enable your account at <https://dashboard.stripe.com/account/international>”.

Changaco:
New error message:

> As per Indian regulations, export transactions require a customer name and address. More info here: https://stripe.com/docs/india-exports

So, we need to modify the payment form to require a postal address when the destination is an Indian Stripe account.

Changaco:
I've just seen another error message:

> As per Indian regulations, only registered Indian businesses (i.e. sole proprietorships, limited liability partnerships and companies, but not individuals) can accept international payments. More info here: https://stripe.com/docs/india-exports

Vonter:
Hey folks. Has there been any payment attempt to an Indian Stripe account that went through successfully?

Changaco:
@Vonter No. (I still haven't modified the payment form to require a postal address when the destination is an Indian Stripe account.)

Changaco:
Collecting postal addresses (#2093) hasn't solved the problem. The latest errors are:

> Stripe doesn't currently support this currency with Amex in India. Refer https://stripe.com/docs/currencies for more details.

> Stripe doesn't currently support destination charges with accounts in IN.

PR #1641, opened 2020-01-01
Update de.po

kensanata:


Changaco:
In the future please submit translations through Weblate instead: https://hosted.weblate.org/projects/liberapay/core/

PR #1642, opened 2020-01-01
pt.po: fix gender of "forma de pagamento"

waldyrious:
I am aware that these files are editable via Weblate, but there seems to be an error with the Portuguese translation at the moment; in the repo there's only a single `pt.po`, but in Weblate there's [pt_PT](https://hosted.weblate.org/languages/pt_PT/liberapay/) and [pt_BR](https://hosted.weblate.org/languages/pt_BR/liberapay/), both of which currently show an error saying:

> The VCS repository has many changes compared to the upstream. Please merge the changes manually or set up push to automate this.

### DIFF
--- a/i18n/core/pt.po
+++ b/i18n/core/pt.po
@@ -1416,7 +1416,7 @@ msgstr "Nova senha"
 
 #, python-brace-format
 msgid "{username} hasn't configured any payment method yet, so your donation cannot actually be processed right now. We will notify you when payment becomes possible."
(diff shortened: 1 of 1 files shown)


Changaco:
The Portuguese translation of the website is in https://hosted.weblate.org/projects/liberapay/core/pt/

waldyrious:
Thanks for pointing me to the right direction. I now realized I clicked "Liberapay Everywhere" instead of "Core" in the main listing at https://hosted.weblate.org/projects/liberapay/. (Maybe because the latter doesn't have "Liberapay" in the name?)

PR #1663, opened 2020-02-11
Fix scheduling

Changaco:
Scheduled PayPal payments shouldn't be tagged as "automatic" since we're not able to execute them automatically.


Changaco:
I used the following code to reschedule payments for the affected donors:

```python
>>> tippers = website.db.all("select distinct tipper_p from scheduled_payins sp join participants tipper_p on tipper_p.id = sp.payer join participants tippee_p on tippee_p.id = (sp.transfers->0->>'tippee_id')::bigint where sp.automatic and json_array_length(sp.transfers) = 1")     
>>> len(tippers)
25
>>> for tipper in tippers: tipper.schedule_renewals()
```

Two of them received `payment_schedule_modified` notifications.

PR #1668, opened 2020-02-16
Improve the notifications page

### LINKED ISSUE #1643: Better UI for deleting notifications and marking them as read
Currently when I open the notifications page I see a button at the top saying "Mark all notifications as read", and each notification has an "X" in the top right, that deletes it if clicked. I clicked on a notification's "X" thinking it would mark it as read, and ended up deleting the notification instead. I think the UI for both features could be improved to avoid similar errors and just to make things clearer overall.

Regarding the "mark as read" feature, it should be possible to mark individual notifications as read from the main notifications page. One way to achieve this is to add a separate icon next to the "X" (say, a checkmark).

As for deleting notifications, IMO that should be better protected against mistakes (including misclicks!). For example:
- It should be clearer what the "X" will do, especially since it's a destructive action. This could be done by adding a tooltip.
- The X could me marked red instead of grey (or turn red on hover)
- There could be a confirmation dialog saying that the action is irreversible
- Even better, the action should be reversible, e.g. instead of disappearing outright, a deleted notification could be replaced with a notice explaining what happened and including a link to undo that action.

Changaco:
Closes #1643.

### DIFF
--- a/liberapay/models/participant.py
+++ b/liberapay/models/participant.py
@@ -1424,8 +1424,10 @@ class Participant(Model, MixinTeam):
     def remove_notification(self, n_id):
         p_id = self.id
         r = self.db.one("""
-            WITH deleted AS (
-                DELETE FROM notifications
+            WITH updated AS (
+                UPDATE notifications
+                   SET web = false
+                     , hide = true
                  WHERE id = %(n_id)s
                    AND participant = %(p_id)s
                    AND web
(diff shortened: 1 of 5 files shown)


Changaco:
Screenshot of some notifications, showing the various states and associated actions

trebmuh:
LGTM

PR #1688, opened 2020-03-02
Add Czechia to Stripe allowed countries

nijel:
Tested on myself.

### DIFF
--- a/liberapay/constants.py
+++ b/liberapay/constants.py
@@ -359,7 +359,7 @@ PAYOUT_COUNTRIES = {
     """.split()),  # https://www.paypal.com/us/webapps/mpp/country-worldwide
 
     'stripe': set("""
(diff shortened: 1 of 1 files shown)


Changaco:
The <https://stripe.com/global> page still says “Preview” for Czechia. We should probably wait a little to see if they update it or if it's going to stay in preview like India (#1606).

nijel:
No problem. I know it's still as a preview, but it now seems to moved to public preview as it now allow registrations while before it didn't.

I accidentally figured out it works when being asked at https://github.com/WeblateOrg/weblate/issues/3540 about enabling IBAN on Liberapay.

Changaco:
The Stripe page hasn't been updated, but payments to Czechia seem to be working fine, so I guess we can merge this.

Issue #992, opened 2018-02-28
PostgreSQL upgrade


Changaco:
[PostgreSQL 10 now Supported in Amazon RDS](https://aws.amazon.com/about-aws/whats-new/2018/02/postgresql-10-now-supported-in-amazon-rds/)

We should migrate our instance from 9.6 to 10, ~~and switch from magnetic storage to SSD at the same time~~ *(this is done)*.

Changaco:
[PostgreSQL 12 has been released](https://www.postgresql.org/docs/12/release-12.html), and [AWS currently supports PostgreSQL 11](https://aws.amazon.com/rds/postgresql/why-postgresql-11/), but we still haven't upgraded from 9.6.

Changaco:
I'm trying this now, hopefully there will only be a few minutes of downtime.

Changaco:
"The instance could not be upgraded because the 'reg*' data type is used in user tables. Please remove all usages of 'reg*' data types and try again."

Changaco:
I'm trying this again. The most recent PostgreSQL version currently supported by AWS is 11.5.

Changaco:
![Screenshot of the AWS database upgrade confirmation page](https://user-images.githubusercontent.com/1581590/76606541-b5139e80-6512-11ea-9a1e-e1d414de8421.png)

Changaco:
The upgrade failed again and caused ~10 minutes of downtime.

```
There were problems executing ""/rdsdbbin/postgres/bin/pg_restore" --host /rdsdbdata/log/upgrade --port 8193 --username rdsadmin --create --exit-on-error --verbose --dbname template1 "pg_upgrade_dump_16390.custom" >> "pg_upgrade_dump_16390.log" 2>&1"
Consult the last few lines of "pg_upgrade_dump_16390.log" for the probable cause of the failure.
```

I don't see the `pg_upgrade_dump_16390.log` file, so I have no idea what went wrong.

Changaco:
The only silver lining is that the app recovered once the DB was back up, so #1662 was a success.

Changaco:
The [AWS Database Migration Service](https://aws.amazon.com/dms/) might work better.

Changaco:
Our reserved database instance is still valid for two months, I'll reconsider what to do then.

Changaco:
I found the error message from yesterday's failed migration while I was looking for something else:

```
Database instance is in a state that cannot be upgraded:
[...]

PR #1704, opened 2020-03-12
Prepare for PostgreSQL upgrade

Changaco:
Unblocks #992 by changing the type of the `statements.search_conf` column from `regconfig` to `text`.


Changaco:
Note: the index recreation produced the following messages 12 times:

```
NOTICE:  word is too long to be indexed
DETAIL:  Words longer than 2047 characters are ignored.
```

PR #1767, opened 2020-05-15
fix: enable users to open url in another tab using ctrl

### LINKED ISSUE #1765: CTRL+click should open in a new tab
Hi,

I was going to trough all the top organizations and picking some out that I wanted to support,
So I was trying to CTRL + click on them to open each one in a new tab, but it ust loads it in the name tab which was annoying
https://liberapay.com/explore/organizations 

Cheers,
Thibault

Laerte:
Close #1765

### DIFF
--- a/js/10-base.js
+++ b/js/10-base.js
@@ -43,9 +43,13 @@ Liberapay.init = function() {
     Liberapay.s3_uploader_init();
     Liberapay.stripe_init();
 
-    $('div[href]').css('cursor', 'pointer').click(function() {
-        location.href = this.getAttribute('href');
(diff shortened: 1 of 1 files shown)


Laerte:
@Changaco Thanks for improvement! Nice catch getting `aux click` event. LGTM.

Changaco:
This is still a dirty hack. I tried a cleaner approach, [redirecting the click events](https://www.growingwiththeweb.com/2016/07/redirecting-dom-events.html) to the link, but the browsers I tested don't act on the cloned events, probably because of anti-popup protections.

PR #1782, opened 2020-05-30
:lock: Corrects shields URLs to utilize encryption

S0AndS0:
This should fix issues of browsers reporting messages similar to...

    Connection
    Connection Is Not Secure
    ⚠️ Parts of this web page are not secure (such as images)

... which pop on _`/<account>/widgets/`_ path.

Hopefully y'all aren't put-off by this PR, because reviewing the HackerOne categories didn't show any that seemed applicable, so I figured a Pull Request would be a swifter fix.

### DIFF
--- a/www/%username/widgets/index.html.spt
+++ b/www/%username/widgets/index.html.spt
@@ -75,20 +75,20 @@ response.headers[b'Content-Security-Policy'] = csp
 
     <p>{{ _("Use these code snippets to add a badge to your website or README:") }}</p>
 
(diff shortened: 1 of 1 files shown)


Changaco:
@S0AndS0 Thank you. For this kind of issue it is indeed better to open a pull request here than an issue on HackerOne.

S0AndS0:
You're very welcome.

I'm glad that this PR was satisfactory in both reporting method and changes submitted. I'll be certain to utilize Pull Requests for anything else that I may find that are quick fixes, and use the HakderOne avenue only if I find anything critical in the future.

Issue #1776, opened 2020-05-20
Unlabeled checkbox when editing profiles.


rkingett:
When editing profiles and accounts, there is an unlabeled checkbox just before the username. I'm not quite sure what it does. I checked it, but nothing seemed to happen. The "profile" checkbox below the username is labeled, but perhaps text should be appended that tells screen reader users a menu opens upon checking this box.

Issue #1775, opened 2020-05-20
Unlabeled avatar header and menu.


rkingett:
When logged in, there is an unlabeled button and an unlabeled graphic. The button opens up a menu, it looks like.

avatar-default etag=.HVo94CoZc…

Right after that is the unlabeled button.

Issue #1774, opened 2020-05-19
Screen reader can't stay focused on languages menu.


rkingett:
Using [NVDA,](https://www.nvaccess.org/about-nvda) an open source screen reader, on [this profile,](https://en.liberapay.com/Liberapay/) for example, NVDA treats the languages menu as one giant level 5 heading, but it is very difficult for the screen reader to stay focused in that languages menu.

I can tell there's a button. In my case, the English button. I'm guessing that's the language currently selected? I wasn't quite sure.

When trying to interact with the language menu on any profile, the screen reader loses focus and jumps around the page wildly.

Perhaps, making the language selections a simple HTML 5 combo box would be the best move.

Changaco:
I don't have any experience with screen readers, but using keyboard navigation I can reach the language links inside the dropdown menu by pressing the down-arrow key twice when the focus is on the button labeled “English” (which is indeed the selected language).

It's possible that dropdowns are more accessible in newer versions of Bootstrap (#940).

Using an HTML `<select>` would be less friendly when JavaScript is disabled.

rkingett:
I had to switch NVDA to [focus mode](https://tink.uk/understanding-screen-reader-interaction-modes/) to tab all the way to the languages menu, the minute I switch back to Browse mode, it causes focus to jump all over the page.

Narrator, a built in screen reader for windows 10, seems to completely ignore the menu when trying to expand that menu in Microsoft Edge and Chrome.

I hope bootstrap becomes more accessible by way of drop down menus. I can attach a video if that would help?

Changaco:
Here's what the documentation for the latest version of Bootstrap says about the accessibility of dropdowns:

> The [WAI ARIA](https://www.w3.org/TR/wai-aria/) standard defines an actual [`role="menu"` widget](https://www.w3.org/WAI/PF/aria/roles#menu), but this is specific to application-like menus which trigger actions or functions. ARIA menus can only contain menu items, checkbox menu items, radio button menu items, radio button groups, and sub-menus.
> 
> Bootstrap’s dropdowns, on the other hand, are designed to be generic and applicable to a variety of situations and markup structures. For instance, it is possible to create dropdowns that contain additional inputs and form controls, such as search fields or login forms. For this reason, Bootstrap does not expect (nor automatically add) any of the `role` and `aria-` attributes required for true ARIA menus. Authors will have to include these more specific attributes themselves.
> 
[...]

Issue #967, opened 2018-01-31
Language localized in their own languages


Jehan:
Currently languages are written in the current localization. For instance, if the page is in English, you would get "French" in the list. But that is a little annoying when you don't speak a language, because then how to find your preferred language in the list?

In a language list, a language must always be self-translated, therefore you will want to see "français" in this list, whatever the lang actually displayed in the rest of the page.

jorgesumle:
> Currently languages are written in the current localization

No they aren't, at least not in the navigation bar. If you mean the language selector from the profile, I don't see nothing wrong with that. Why would you want to (select and) write a statement in a language you don't understand? If you can please specify a bit more what do you refer to, it would be better.

Jehan:
> No they aren't, at least not in the navigation bar.

I'm not sure what you are refering at, I don't see any language list in the navigation bar here (I tried every item in the nav bar just now to be sure). :-)

Personally I simply meant the list of language for a project description. Here is the screenshot for our project for instance (which has English, French and Korean; I would expect to see the list: English, français, 한국어):
![screenshot from 2018-02-01 01-21-06](https://user-images.githubusercontent.com/794997/35654824-3095c47c-06ef-11e8-9d18-db0eb3446d15.png)

> Why would you want to (select and) write a statement in a language you don't understand?

I'm not sure what you mean about writing in a language that you don't understand (this seems to be like impossible by definition). But I can answer about how would you arrive on the localization you don't understand, there are many possibilities for this: maybe you just followed a link which was directing to a specific lang (ko.liberapay.com/…); maybe you followed the generic link, but being in a foreign country, the language detection gave you the localization for this country; maybe the computer you are in right now is set for this language (company computer, friend's, whatever…) even though not your favorite lang; maybe you speak/read this other language and have your computer set on purpose for it, yet you still prefer to read in some other language sometimes…

Also why would you want to get each language self-translated? Simply to be able to find yours easily. If I look for the French language, I will find it a lot easier if I see "français" in the list, rather than "프랑스어".
And this, even more if I happen to not speak Korean and came into the Korean version by mistake. 
Then it's quite hard to know which is which language (and even simply to be sure this is the language list at all, except by trial and error). :-P

As an example, here is the language list selection for GIMP (it's a known example, I am the one who made the languages self-translated there ;p):
[...]

PR #1784, opened 2020-06-01
Improve accessibility

Changaco:
This branch fixes a bunch of issues, including #967, #1774, #1775 and #1776.

### DIFF
--- a/js/10-base.js
+++ b/js/10-base.js
@@ -63,7 +63,7 @@ Liberapay.init = function() {
         }
     });
 
-    $('.navbar .dropdown-hover').removeClass('dropdown-hover');
+    $('.dropdown.dropdown-hover').removeClass('dropdown-hover');
 
     $('.dropdown-toggle-form').click(function() {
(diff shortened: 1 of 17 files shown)


rkingett:
Is there any way the below can be done without using ARIA? ARIA should never be used if it can be achieved natively.

> add an extra ARIA label to subnav checkboxes

Could you do this natively without ARIA?

Changaco:
I guess it's technically possible to add a second `<label>` specifically for screen readers, instead of an `aria-label` attribute on the checkbox, but would that really be any better? I don't think so.

These checkboxes aren't visible on screen, their purpose is to provide a toggling mechanism based on CSS instead of JavaScript.

rkingett:
Gotcha! I thought they were visible on screen. On second thought, ARIA would work in this particular instance.

Changaco:
These changes are now deployed. @rkingett If the result is unsatisfactory, or if you find other problems, you can of course open new issues.

PR #1804, opened 2020-06-25
Add donation form validation of Custom amounts (liberapay#1696)

### LINKED ISSUE #1696: Warn when custom amount is invalid
Steps to replicate
------
1. Click "Donate" on any project's page
2. Choose "Custom"
3. Don't input any value in the "amount" field or input an invalid value (e.g. "abc").
4. Click "Donate" 

Current behavior
-----
The page "Bad Request" appears, saying '"" is not a number'.

Desired behavior
-----
After clicking "Donate", the "Amount" field should be highlighted in red, with an error message shown right below the field.

1124816:
Closes #1696. 
This is my first contribution so pointers very welcome. I wasn't sure how best to make the error message translatable so I made up an html attribute.

### DIFF
--- a/js/10-base.js
+++ b/js/10-base.js
@@ -100,6 +100,7 @@ Liberapay.init = function() {
         $this.parents('form').find('input').on('change', function() {
             $this.prop('required', $requirer.prop('checked'));
         });
+        $requirer.trigger('change');
     });
 
     $('[data-toggle="tooltip"]').tooltip();
--- a/js/10-base.js
+++ b/js/10-base.js
@@ -93,17 +93,24 @@ Liberapay.init = function() {
         ).prop('selected', true);
         history.pushState(null, null, location.pathname + href + location.hash);
     });
+    
+    $('input[data-required-if-checked]').each(function(index, value) {
+        var radio = $(value).attr('data-required-if-checked');
+        $($(radio).parents('form').find('input')).change(function() {
+            $(value).prop('required', $(radio).is(':checked'));
+        });
+    });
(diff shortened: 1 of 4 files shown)


1124816:
I think this is what you are asking for? Should I squash the commits?

js/10-base.js
```diff
@@ -93,6 +93,17 @@ Liberapay.init = function() {
         ).prop('selected', true);
         history.pushState(null, null, location.pathname + href + location.hash);
     });
+    
+    $('input[data-required-if-checked]').each(function(index, value) {
+        var radio = $(value).attr('data-required-if-checked');
+        $($(radio).parents('form')).on('click', function(e) {
```
Changaco, line 99: Clicking isn't the only way to interact with a form. The proper solution is probably to listen for the [`change`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event) or [`input`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event) event on the radio input(s).

js/10-base.js
```diff
@@ -93,6 +93,17 @@ Liberapay.init = function() {
...
+            if($(radio).is(':checked')) {
+                $(value).prop('required', true);
+            } else {
+                $(value).prop('required', false);
+            }
```
Changaco, line 104: This `if … else …` block can be simplified to `$(value).prop('required', $(radio).is(':checked'));`.

1124816:
Thanks!

Issue #1287, opened 2018-10-21
Fees


Changaco:
With Mangopay we computed the fees and added them on top of how much a donor wanted to put into their wallet, but with Stripe and PayPal we're unable to predict the fees because they're too complex, so when a donor sends 4 weeks worth of a donation the recipient actually gets 3.x weeks worth, and because of how the accounting currently works the donor is asked to send money again 3 weeks later instead of 4 weeks later. This is confusing and needs to be fixed.

- [ ] modify the code that tracks how much has been paid in advance to use the amount sent by the donor instead of the amount received by the donee
- [ ] document how fees work

MartinDelille:
I was asking myself how much were stripe fees and I found this: https://stripe.com/pricing

tuxayo:
Would it be easier to show an estimate with Stripe direct debits? For example I'm donating from eurozone 1€/mo and paying one year in advance. That's a 12€ payment.

According to https://stripe.com/fr/pricing
>  SEPA Direct Debit 
>  0.8% + €0.25
> €5 cap 

12*0.08+0.25 = 1.21 (10% fee, it's 3 times the fees that I have with Stipe credit card payments, so for my amount it's not the one I should choose)

tuxayo:
For anyone wondering: here are real examples of fees that I get from France, euro to euro.
```
- 12,00€ Stripe credit card → € 0,42 fees so 11,58 € donated: 3.5% (0.42/12*100)
- 12,00€ PayPal             → € 0,76 fees so 11,24 € donated: 6.3% (0.76/12*100)
```
Maybe I'm wrongly calculating fees: (0.42/12*100) vs (0.42/(12-0.42)*100)

Changaco:
> Do you think it's by design that there's no way to query the PayPal/Stripe API to determine the fee before charging?

It's because businesses don't care what the exact fee is, they don't upcharge customers, and most platforms take their own cut so the payment processor's fee is included in that.

mattbk:
Relevant? https://support.stripe.com/questions/what-is-the-scope-of-the-surcharge-ban-under-psd2

P4E:
> Would it be easier to show an estimate with Stripe direct debits? For example I'm donating from eurozone 1€/mo and paying one year in advance. That's a 12€ payment.
> 
> According to https://stripe.com/fr/pricing
> 
> > SEPA Direct Debit
> > 0.8% + €0.25
> > €5 cap
> 
[...]

PR #1820, opened 2020-07-23
Send renewal reminders later in some cases

### LINKED ISSUE #1818: Liberapay asks to renew a donation even though it's not time yet
image

The title says it. I've received an email asking to renew the donation even though I've just started it a week ago and the payment already got through. Clicking the green button promts to pay the donation on the picture, even though it says "due in two weeks".

Changaco:
Closes #1818.

### DIFF
--- a/liberapay/payin/cron.py
+++ b/liberapay/payin/cron.py
@@ -25,7 +25,19 @@ def send_donation_reminder_notifications():
                    SELECT sp.id, sp.execution_date, sp.amount, sp.transfers
                ) a ORDER BY a.execution_date)) AS payins
           FROM scheduled_payins sp
-         WHERE sp.execution_date <= (current_date + interval '14 days')
+     LEFT JOIN LATERAL (
+                   SELECT pi.*
+                     FROM payins pi
+                    WHERE pi.payer = sp.payer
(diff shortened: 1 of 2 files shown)


mattkwarren:
I may be misunderstanding but this issue and #1818 do not appear to be fixed. See attachment. This is really confusing UX if it isn't a bug. The renewal information certainly is not correct. A monthly payment started 11 minutes ago should not be due in 2 weeks.

liberapay

Changaco:
@mattkwarren Your problem is #1287.

mattkwarren:
I don't think it is my problem. It is a Liberapay UX problem that has existed since what looks like at least 2018. Issue #1287 doesn't appear to indicate why the card charge date displayed in my view needs to be tied to fees or what recipients will receive. I'm completely missing why the UX needs to be this way. Payment processor fees and donation amounts don't seem to have anything to do with when my card will be billed again, as I authorized monthly (not bi-weekly). Current UX is confusing as a user/donor and really should be changed. I don't find any of the explanations sufficient. Although, that's just my input. I planned to donate to Liberapay but have no intention of contributing to the codebase.


Feb 5, 2023, 04:34 by ***@***.***:

PR #1821, opened 2020-07-23
New daily cron job to fix inconsistencies in scheduled payins

Changaco:
In some cases scheduled payins can be missing or inconsistent, particularly after a creator adds or removes a payment account.

### DIFF
--- a/liberapay/cron.py
+++ b/liberapay/cron.py
@@ -22,13 +22,13 @@ class Cron(object):
         self.website = website
         self.conn = None
         self.has_lock = False
-        self.exclusive_jobs = []
+        self.jobs = []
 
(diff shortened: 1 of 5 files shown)


Changaco:
Deploying this broke the app. There are many `PoolError("connection pool exhausted")` exceptions in Sentry; for some reason I wasn't able to SSH into the server (`kex_exchange_identification: read: Connection reset by peer`); restarting the app server through the AWS EB dashboard didn't help; and re-deploying the previous version of the app didn't work (AWS got stuck for a while, which is what tends to happen when something goes wrong, because of the lack of properly short timeouts on operations). The deployment of a new instance has now succeeded, so the app is back up.

Issue #586, opened 2017-05-02
Tying releases to commits in Sentry


Changaco:
[Sentry has improved its GitHub integration](https://blog.sentry.io/2017/05/01/release-commits.html). I've set up the webhook on GitHub, but to complete the integration we need to tell Sentry about our releases. That probably means two things:

- [ ] calling its API with `curl` from our `release.sh` ~~or [from Travis](https://docs.sentry.io/workflow/releases/release-automation/travis-ci/)~~ *(can't work because releases are tagged after Travis runs)*
- [x] adding a `release` argument to the `raven.Client()` call in `wireup.py` (need to figure out how to get the release number) *(done in #929)*

Changaco:
https://docs.sentry.io/clients/python/advanced/#configuring-the-client

Changaco:
We should also improve the tags we attach to events: <https://docs.sentry.io/learn/context/>. In particular we already send `user_id` and `username` but we don't do it the right way, so Sentry doesn't recognize them as user tags.

Changaco:
> and adding a `release` argument to the `raven.Client()` call in `wireup.py` (need to figure out how to get the release number)

AWS EB doesn't seem to provide easy access to the current app version number to the application itself. As far as I can tell it's only available as a key in the obscure JSON file `/opt/elasticbeanstalk/deploy/manifest`. Example in which `258` is the version number:

```
{"RuntimeSources":{"liberapay.com":{"258":{"s3url":""}}},"DeploymentId":216,"Serial":266}
```

Changaco:
[Automate Release Management with the Sentry Release GitHub Action](https://blog.sentry.io/2020/07/30/automate-release-management-with-the-sentry-release-github-action)

PR #1834, opened 2020-07-31
Tell Sentry about releases

Changaco:
This branch is an attempt to implement #586.


Changaco:
I've just realized that this can't work, because a release isn't tagged yet when Travis tests the commit.

PR #1849, opened 2020-08-11
New log-in page for donors

Changaco:
When an unauthenticated visitor submits a donation form, they see a page asking them to log in or create an account ([screenshot](https://user-images.githubusercontent.com/1581590/89876160-8f242000-dbbe-11ea-993f-865a077e11d9.png)). That page is a little overwhelming and disturbs some people, particularly those who dislike the idea of having to create an account, so I'm proposing a simpler page that doesn't use the word “account”.

Screenshot of the new log-in page for donors

---

Screencast of the new donation flow for a new donor


rkingett:
We should also make sure this new page conforms to accessibility standards and inclusive design.

Changaco:
The new page is simple HTML, like the current one, so I don't expect any accessibility issue.

Changaco:
A remaining problem with this new page is that the donor is asked to accept Liberapay's terms of service. Removing this distraction requires further consideration and work.

Changaco:
A way to contact the payer is a necessity for a recurrent donation, we need to be able to notify them when it's time to renew the donation. Email addresses are the standard contact method, and they're stable identifiers that enable grouping a person's donations together into a single account.

Changaco:
Setting up a standing order through your bank's website might seem simpler to you, but it's a payment method that we can't support, because it's too problematic.

Bubu:
Hey, @changaco, sorry to bother you about this some more but we've been asked a lot why liberapay doesn't "just" support this method of donations. Could you elaborate a bit here why this is problematic? This would help pointing people to when this comes up again. :)

Changaco:
@Bubu See <https://github.com/liberapay/liberapay.com/issues/2026>.

Cher172:
What?

PR #1895, opened 2020-10-04
Fix 'mark notifications as read' message

waldyrious:
Only the notifications from the current page are marked as read, not all of the unread ones, as the message previously implied.

### DIFF
--- a/i18n/core/ar.po
+++ b/i18n/core/ar.po
@@ -3290,7 +3290,7 @@ msgstr ""
 msgid "Please confirm"
 msgstr ""
 
-msgid "Mark all notifications as read"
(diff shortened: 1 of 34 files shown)


waldyrious:
Note: I've edited the translation files accordingly, but if that's not the correct procedure to update them, I can remove those changes from the commit.

Changaco:
I think we should fix the behavior of the button instead of changing its label. I've written a patch for this, I will push it later in another branch.

waldyrious:
Ah, ok. I thought the behavior was intentional :) Feel free to close this, then.

Changaco:
Done in #1897.

PR #1920, opened 2020-11-19
Require password to add a new email address

Changaco:
This change prevents someone who has access to a user's session, but not to that user's password, from taking over the account.


baibhavanand:
yay 🎉 I believe it was one of my first bugs on HackerOne. Though it took a long to get resolved I am happy to be able to give my input on Libra Pay.

PR #1951, opened 2021-01-10
Reword

Changaco:
This branch modifies the FAQs below donation forms, and replaces the verb “stop” with “discontinue” for recurrent donations, because it sounds better to me.

### DIFF
--- a/www/%username/donate.spt
+++ b/www/%username/donate.spt
@@ -125,16 +125,26 @@ full_title = _("Donate to {0} via Liberapay", participant.username)
 
     <h4>{{ _("Can I make a one-time donation?") }}</h4>
     <p>{{ _(
-        "No, but you can stop your donation at any time, and it will de facto "
-        "cease if you don't top up your account when needed."
+        "One-time donations aren't properly supported yet, but you can discontinue "
+        "your donation immediately after the first payment."
     ) }}</p>
 
-    <h4>{{ _("Is this platform legit and secure?") }}</h4>
+    <h4>{{ _("What is this website? I don't recognize it.") }}</h4>
     <p>{{ _(
-        "Yes. Liberapay is run by a non-profit organisation {0}legally registered{1} "
-        "in France, we comply with the European Union's financial regulations, and "
-        "the security of our service is of paramount importance to us.",
-        '<a href="/about/legal">'|safe, '</a>'|safe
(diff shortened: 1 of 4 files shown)


www/%username/donate.spt
```diff
@@ -125,16 +125,21 @@ full_title = _("Donate to {0} via Liberapay", participant.username)
...
-        "cease if you don't top up your account when needed."
+        "One-time donations aren't properly supported yet, but you can discontinue "
+        "your donation immediately after the first payment."
     ) }}</p>
 
-    <h4>{{ _("Is this platform legit and secure?") }}</h4>
+    <h4>{{ _("What is this website? I don't recognize it.") }}</h4>
     <p>{{ _(
-        "Yes. Liberapay is run by a non-profit organisation {0}legally registered{1} "
-        "in France, we comply with the European Union's financial regulations, and "
```
trebmuh, line 135: > we comply with the European Union's financial regulations

This is informative and reassuring for readers. It probably should stay here then.

Yetizone:
Hello, newcomer here interested in open source crowdfunding platforms. I recently discovered Liberapay and wanted to try contributing. 

I'll start by giving my 2 cents on this PR: 
* I agree with the rewording from "stop" to "discontinue"
* I also strongly agree with the rewording around the one-time donation FAQ, I think the new proposed answer is definetly better :) 
* Lastly, I disagree with the split & reword of the last FAQ, the one about the platform being legit & secure. I find the current FAQ and answer to be well worded, reassuring and informative and I think they should be left as-is. Most importantly, I think the link to `about/legal` and the statement about complying to EU's regulations should be kept in the answer. 

Of course, this was just my opinion, but I hope it will be helpful nonetheless :) 

PS: Maybe the last FAQ could be changed to something like "Is this platform secure ?", i.e. to remove the "legit" part, as legit doesn't have a very clear meaning in this context and I'm also not sure if it's easily translatable. However I would still find the current answer to be very good as-is :)

Yetizone:
PS part 2: Another option might be to expand "legit" into "legitimate", this could make the question easier to translate and also give it a more serious tone. However I still think the simpler and better option is to just have the question be "Is this platform secure ?".

Alright, that was my final rambling on wording :)

Changaco:
I have added back a mention of EU laws, but not the link to the `/about/legal` page, because I think that it's probably more distracting than reassuring.

PR #1963, opened 2021-01-23
Upgrade to Bootstrap 5 (WIP)

CrafterSvK:


CrafterSvK:
Libsass is depracated

CrafterSvK:
LibSass is depracated. If we want to use newer features of SASS/SCSS we should move to DartSass not sure about python tho.

PR #1969, opened 2021-01-29
Don't send partially translated emails

Changaco:
I think it's probably better to fall back to English completely instead.


trebmuh:
It makes sense.

PR #1991, opened 2021-03-23
Email log-in confirmation

Changaco:
This branch adds a confirmation step when logging in via email. The goal is to make it more difficult for an attacker to trick a target into using an account controlled by the attacker.


Changaco:
Screenshot of the confirmation message:

“You are about to log in as `someone@example.com`. Please click a button below to confirm or cancel.”

leojoy10:
Hi @Changaco,

Thanks for the reply. 
Yes, This is the correct mitigation to implement if another user is about to login in the same browser

Issue #1932, opened 2020-12-14
Accepting or refusing an invitiation to join a team can't be reversed


Changaco:
People can make mistakes and change their minds, it's not good to hinder that. The code in `www/%username/membership/%action.spt` should be modified to allow the invitee to accept an invitation after having refused it, and to refuse an invitation after having accepted it (with a warning message asking them to confirm that they want to leave the team).

evcheng1:
I plan on working on this and was wondering, since users can leave a team, what is the  additional benefit of allowing them to refuse an invitation after having accepted it? Isn't that functionally the same thing as leaving the team? Thank you!

Changaco:
> what is the additional benefit of allowing them to refuse an invitation after having accepted it?

Avoiding user confusion and frustration. When a user tries to reject and invitation after having mistakenly accepted it, seeing the error message “You have already refused this invitation” isn't very helpful.

> For example, would there be some sort of separate page of past rejected invitations that a user can visit?

That's a separate issue: #590.

> And are these rejected invitations always available to be accepted, or should they expire?

Invitations should probably expire, but currently they don't. You can implement that too if you want, but it's not a requirement to close this issue.

PR #1997, opened 2021-04-16
Reverse accepting or refusing a team invitation

evcheng1:
Users can reverse their decision in accepting or refusing a team invitation (addresses issue #1932).


Changaco (changes requested):
I think the `leave_team` function should be a method of the `Participant` class and shouldn't handle the confirmation.

Changaco (approved):

Issue #463, opened 2016-12-22
Income who are not easy divisible are wrong


Zatalyz:
A little bug, but it can be problematic in team. Excellent example here : https://en.liberapay.com/Khaganat/income/

Team have 0,85€/week and 7 people. It divide in 7 ( 0,121428 etc), rounded to the decimal point top (0,13), and the last member have less (0,85-(6*0,13)=0.07). Because people are alphabetical ordered, it's for me ;) but, also, it's always the same people who haven't got all of his part. 

Better to rounded to the decimal point lower and so have "Leftover". And when leftover is divisible by the number of people, it's shared. 

Example : like now, Team have 0,85€/week and 7 people. 0.85/7 and rounded lower = 0,12, rest 1 centime. So leftover have 1 centime per week ; after 7 week, if they are always 7 people in the team, all people have 1 centime in addition. And all people have the same income if all have the same part :)

Changaco:
Originally this was supposed to be solved by processing team takes in random order, so it wouldn't be the same member who gets less each time, but this is linked to #434 and I haven't gotten around to that yet.

Keeping track of unused team income is also a good idea though. I've been planning to implement it as well, but more for #81.

lewjs16:
I was wondering if this issue was still open & needed to be fixed. I am interested in working on it/looking into it further!

Changaco:
I think this can still happen. I recommend adding a test in the `tests/py/test_payday.py` file to reproduce the problem: a team with €0.85 income and 7 members each taking €1. This should confirm that the `Payday.resolve_takes()` function needs to be improved.

The issue can probably be fixed by replacing the simplistic `takes_ratio` ([payday.py#L472-L477](https://github.com/liberapay/liberapay.com/blob/593/liberapay/billing/payday.py#L472-L477)) with a call to the generic `resolve_amounts()` function ([common.py#L493](https://github.com/liberapay/liberapay.com/blob/593/liberapay/payin/common.py#L493)).

mhogara:
We’ve been working on this issue and have successfully made a test to reproduce the problem and have it failing on the current version of the code. 

Fixing the issue itself has been a bit more difficult though; we are trying to call `resolve_amounts()`(common.py#L493), but are not sure what exact arguments to pass in. As of right now, we think that `available_amount` would be `fuzzy_income_sum` and `convergence_amounts` could be a dictionary of IDs or indices and corresponding takes. However, we aren’t sure what should be passed in for `base_amounts`. 

[...]

PR #2000, opened 2021-04-20
Resolving Issue #463: Income is not easily divisible

mhogara:
**Problem:** As described in issue 463, previously, incorrect rounding could lead to one single team member getting a significantly different amount each week. 

**Solution:** We changed a call (within resolve_takes) to resolve_amounts to generally better distribute takes. We also made a specific if statement that checks for the specific case described in the original issue report and randomly selects a team member to get one single extra cent each week. 
The random selection required importing the random library and adding a large section of code in resolve_amounts. An alternative solution could be giving the same team member one extra cent every week. This is still an improvement on the original problem as the net difference between takes is only the minimum currency value(i.e. one cent). Let us know in reviews if this non randomized version is a preferred solution. 

**Overview of changes:** Added test in test_payday, added call in payday.py, changed values in resolve_takes to match currencies, randomized extra cent distribution in resolve_amounts

### DIFF
--- a/liberapay/billing/payday.py
+++ b/liberapay/billing/payday.py
@@ -11,7 +11,7 @@ from babel.dates import format_timedelta
 import mangopay
 import pando.utils
 import requests
-
+from liberapay.payin.common import resolve_amounts
 from liberapay import constants
 from liberapay.billing.transactions import Money, transfer
 from liberapay.exceptions import NegativeBalance
@@ -599,7 +599,7 @@ class Payday:
                     take.amount -= transfer_amount.convert(take.amount.currency)
                 if in_advance_amount:
                     tip.paid_in_advance -= in_advance_amount
-                    take.paid_in_advance -= in_advance_amount.convert(take.amount.currency)
+                    take.paid_in_advance -= in_advance_amount.convert(take.paid_in_advance.currency)
                 if on_time_amount:
                     tip.balances -= on_time_amount
                 tip.amount -= transfer_amount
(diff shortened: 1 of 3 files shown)


mhogara:
Saw that this failed the Travis CI build- will work on fixing those bugs shortly, but it appears to be mainly formatting errors.

Changaco (changes requested):
The member who gets the extra cent shouldn't be selected randomly, because that's unstable. A relatively simple solution is to use the ID of the ~~latest or~~ next payday instead.

Also, please preserve the formatting of the code you're modifying, including the grouping and ordering of the imports.

mhogara:
An update: We have successfully fixed all the formatting errors and tried to match original imports as much as possible.

In terms of using the ID of the latest to distribute the extra cent, does that mean the ID of the next member of the team?  In the case of our test file, this would be the "jillian" each time. If this is not what you mean, please elaborate on what "ID of latest or next payday" is referring to so we can make sure our implementation matches your review.

Changaco:
The ID of the next payday is basically the number of weeks since Liberapay was launched. Right now it's 272, and later today it will be 273.

This number will have to be passed to the `Payday.resolve_takes` method as a `payday_id` argument. It can be obtained like this:

```python
next_payday_id = cursor.one("""
    SELECT id
      FROM paydays
     WHERE ts_start IS NOT NULL
  ORDER BY id DESC
     LIMIT 1
""", default=0) + 1
```

mhogara:
Alright, so to be clear in the proposed solution we would use ID of the next payday mod-ed by the number of team members to choose who gets the extra cent- that way it would change every week and cycle through each member. Let me know if this understanding is correct.

Changaco:
@mhogara You understood correctly.

Changaco:
@mhogara Are you still working on this?

mhogara:
Yes we are; sorry for the delay, we were busy with final exams for a few weeks, but do plan on finishing this up soon.

Changaco:
@lewjs16 @mhogara No worries. Good luck with your new jobs. :slightly_smiling_face:

PR #2053, opened 2021-07-29
No self disavowal

Changaco:
This branch tries to prevent users from blocking their own accounts.

### DIFF
--- a/liberapay/models/participant.py
+++ b/liberapay/models/participant.py
@@ -1820,7 +1820,7 @@ class Participant(Model, MixinTeam):
         host = website.canonical_host
         username = self.username
         if query:
-            query = '?' + urlencode(query)
+            query = '?' + urlencode(query, doseq=True)
         email_row = getattr(self, '_rendering_email_to', None)
         if email_row:
             extra_query = []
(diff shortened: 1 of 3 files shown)


Changaco:
For the record, I've used this script to unblock the accounts that were self-disavowed in the past:

```sql
BEGIN;
    CREATE TEMP TABLE self_disavowed ON COMMIT DROP AS
    SELECT em.*
      FROM emails em
      JOIN events ev ON ev.participant = em.participant
                    AND ev.type = 'email_disavowed'
                    AND ev.payload->>'address' = em.address
     WHERE em.disavowed
       AND ev.recorder = em.participant;

    UPDATE emails
       SET disavowed = false
         , verified_time = current_timestamp
         , verified = true
     WHERE id IN (SELECT id FROM self_disavowed);

    UPDATE email_blacklist
       SET ignore_after = current_timestamp, ignored_by = 1
     WHERE lower(address) IN (SELECT lower(address) FROM self_disavowed);

    UPDATE participants AS p
       SET email = (
               SELECT e.address
                 FROM emails e
                WHERE e.participant = p.id
                  AND e.verified IS true
             ORDER BY e.added_time
                LIMIT 1
           )
     WHERE p.id IN (SELECT participant FROM self_disavowed)
       AND coalesce(p.email, '') <> coalesce((
               SELECT e.address
                 FROM emails e
                WHERE e.participant = p.id
                  AND e.verified IS true
             ORDER BY e.added_time
                LIMIT 1
           ), '');
END;
```

Output:

```sql
BEGIN
SELECT 62
UPDATE 62
UPDATE 12
UPDATE 62
COMMIT
```

PR #2062, opened 2021-08-22
Read-only sessions

### LINKED ISSUE #261: Implement 'keep me logged in'

Changaco:
This branch changes how user sessions work. Instead of expiring after 6 hours of inactivity, a session will be downgraded to read-only 6 hours after its creation, and that read-only session will expire after 4 weeks of inactivity.

Additionally, the links in email notifications sent to the owners of accounts lacking a password will include session tokens, so that clicking on them will open a full read-write email session if within 6 hours, and a read-only session if within 4 weeks.

The session cookies will have an expiration date ten years in the future instead of only 6 hours, not because they're valid for ten years, but because receiving an expired session cookie tells us that the visitor has most likely used Liberapay before.

Closes #261, without adding any “keep me logged in” checkbox.


Changaco:
Screenshot of a reauthentication form

PR #2083, opened 2021-11-01
autofill elsewhere if we have a query

AngryUbuntuNerd:
I just noticed that on the Search page, the query does not get repeated in the "Didn't find who you were looking for?" part, so you have to copy-paste it again.

This is a quick untested fix based on what I saw in the code base.

Could someone check it quickly?

### DIFF
--- a/templates/macros/elsewhere.html
+++ b/templates/macros/elsewhere.html
@@ -61,7 +61,7 @@
 % macro user_lookup_form()
     <form action="/on/" class="form-group">
     <div class="form-inline">
-        <input name="user_name" placeholder="{{ _('username') }}" autocorrect="off"
+        <input name="user_name" placeholder="{{ _('username') }}" value="{{ query or '' }}" autocorrect="off"
                spellcheck="false" size="12" type="text" class="form-control">
         <select class="form-control" name="platform">
             % for platform in website.platforms.hasattr('x_user_name')


jorgesumle:
You are assuming that people will always search a username, but they could search by project name or project description as well.

AngryUbuntuNerd:
they can still change the query, but now its not blank but prefilled if they wish to search for the same (which I personally did all the time)

jorgesumle:
But prefilling isn't good if they didn't type a username, because they would have to delete the prefilled text before entering the new text, and they wouldn't be able to see the placeholder.

AngryUbuntuNerd:
okay I think there is two issues here

1. this has not been communicated very well. closing this ticket in less than 24h with two vaguely expressed opinions is merely a decision process. it doesnt shed a good light on encouraging contributions.

2. the way I understood your viewpoint does not seem to be a good trade off. so lets look at it in detail from my viewpoint:
* I click Search, a form field opens
* I enter a project I like, i.e. "kodi" (possibly dont yet care if its a username or whatnot)
* I see no results
* at the bottom of the page it says "Didn't find who you were looking for?" and a form field
* but its not filled with anything. now I have to copy-paste the search term into the same form field.
* SUGGESTED CHANGE: that form field is now prefilled - no more copy-paste necessary

* YOUR ISSUE: I am looking for something that is not a username. that assumes I know that I am looking for a username first. and secondly if I know about that issue, the solution is merely to double click the form field - or to tab into it, both will highlight the text, so I can enter anew. thats less work than copy-pasting, and also your scenario (specifically knowing to search for a username) is imho less likely. how is that little tradeoff worth accepting the UI inconvenience of not prefilling?

Issue #972, opened 2018-02-05
Unable to disambiguate "Donate" string at Weblate


jmontane:
Hi,

AFAIK, "Donate" string is used at 3 different places, but [a single string](https://hosted.weblate.org/translate/liberapay/core/ca/?checksum=fa0b00bbfa206744) is available for translating "Donate" at Weblate only.

I request adding context support for this string (for all strings better?), so different translations could be provided.

- Donate button
![donate-1](https://user-images.githubusercontent.com/339645/35804845-55c09a9e-0a71-11e8-8491-e1a5f5193886.png)

- Donate widget button
![donate-3](https://user-images.githubusercontent.com/339645/35804843-557d1fc6-0a71-11e8-9166-b9c98ead54f0.png)

- Donate header
![donate-2](https://user-images.githubusercontent.com/339645/35804844-559ea7ae-0a71-11e8-8f4f-1bb67373e6d7.png)

PR #2087, opened 2021-11-12
WIP: [i18] Add context to 'donate'

jorgesumle:
Aims to solve #972

It's a matter of [using contexts](https://www.gnu.org/software/gettext/manual/html_node/Contexts.html) where necessary. I think it would be reasonable to reuse the existing translation of "donate" (in languages where it's translated) for these new strings (because these 3 new contexts mean new untranslated `msgid`s in the `.po` file).

### DIFF
--- a/www/data/exchange-rates.spt
+++ b/www/data/exchange-rates.spt
@@ -1,0 +1,13 @@
+[---]
+rates = dict(website.db.all("""
+    SELECT target_currency, rate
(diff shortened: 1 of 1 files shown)


Changaco:
For this to work the `pgettext` function would have to be implemented in `liberapay/i18n/base.py` and injected into the context along with `_` and `ngettext`.

Issue #2013, opened 2021-04-30
Missing profile descriptions


Changaco:
We have too many users who receive payments but don't have any profile description. We should probably send them a notification asking them to complete their profile and warning them that an incomplete profile increases the probability of their account being blocked.

Urahara:
The reason of account being block is suspicious activty?

Changaco:
@Urahara Suspicious payments, yes.

sjcummins:
Hi, I am interested in working on this issue if it is still available

Changaco:
Yes, this issue still available.

sjcummins:
Just a few questions--
1. How should we define an empty description? Just an empty string? One that doesn't contain any alphanumeric characters (e.g. a bunch of spaces)?
2. How long should we wait to send the reminder (i.e. x days after the account receives its first payment)?
3. Would having a short description or a full description enough to fulfill the requirement or is both needed?
4. Should the notification be repeated after a certain amount of time if there is still no description? If so what should this period of time be?
5. What should the notification message be?

Thanks in advance for the clarifications

Changaco:
1. We don't need anything complicated, just check whether the description exists.
2. I'm not sure why this notification should be delayed at all, but when in doubt I use 6 hours as the delay.
3. Only the full description should be required.
4. I guess it would be okay to send it every 6 months, but only if there have been new payments received since the previous notification. It would also be okay to send it only once.
5. I don't know what the exact message should be.

bearbin:
The wording of the new notification is rather strong, particularly since it is being sent to users who have had an account and been receiving payments for some time.

The warning about account disablement should be qualified by a statement along the lines of "in the case of suspicious payments", and a link to the moderation policies added.

PR #2096, opened 2021-12-08
Notify users who don't have any profile description

sjcummins:
Implemented Issue #2013 to send a notification to users without a profile description on their first payday encouraging them to add a profile description. The notification is sent at the same time as the income notification and is only sent once (first payday). The email informs them why they should add a description (suspicious/fraudulent activity) and a link to edit it. Testing includes creating users with and without profile descriptions and payment accounts and ensures those that successfully receive a payment for the first time receive the add_profile_description notification as well, and those with a description or without a payment account do not receive the notification.


Changaco (changes requested):
- [x] The new notification should be sent to all users who have received money and don't have a profile description, not only to users who will start receiving in the future. A daily or weekly cron job would probably be more appropriate than doing this as part of payday.
- [ ] It would be better if the new notification's name was more consistent with the names of the existing notifications. It should probably be `profile_description_missing`.

sjcummins:
For setting it up as a cron job, I'm struggling to think of an efficient way to keep track of who has already received the notification so they don't receive a new one every time it's run. Sending during the payday allowed me to use the data from the generate_income_notifications queries which made it so I only had to consider users who received income that week and never received income before that.

For what you're saying, I think I'd have to store a boolean value for each user indicating whether they already got notified, and the only way I can think of to do that is to modify the database schema. Is that the best way to do it?

Changaco:
To determine if the notification has already been sent to a user, simply look for it in the `notifications` table.

sjcummins:
^^ let me know if the cron job should be somewhere else; I wasn't really sure where to put it but I didn't think it should stay in payday.py

Changaco:
You can put the cron job function with the existing ones at the end of the `liberapay/models/participant.py` file.

sjcummins:
I think it should be good now; let me know if there’s anything else that needs to be changed.

Changaco:
I've merged this manually because GitHub wouldn't let me force-push the branch.

Changaco:
@sjcummins Congratulations on your first contribution, and sorry it took longer than you were hoping for.

trebmuh:
:+1:

Issue #1965, opened 2021-01-27
NSFW flag for accounts


Changaco:
Liberapay profiles probably need a “Not Safe For Work” flag that can be set by profile owners and Liberapay admins, similar to the privacy settings but for a different purpose.

Changaco:
The database schema changes and admin part of this has been implemented in #2024. I've used the word “unsettling” instead of the abbreviation NSFW in the code and the admin interface.

What's left to do:

- [ ] Enable users to mark their profiles as NSFW. I suggest renaming the existing `/~/edit/privacy` page to `/~/edit/visibility` and putting the new checkbox there.
- [ ] Hide profiles marked as unsettling, unless the viewer has opted in to see them.

longj724:
I am looking to contribute to the repository for the first time. Does this issue still need to be worked on?

Changaco:
Yes.

longj724:
I've looked into this task more and am interested in working on it. I understand where users should be able to set their profile as "unsettling", but where users be able to select to opt in to viewing "unsettling" profiles?

Changaco:
On every page which would show an unsettling profile if the user had already opted-in to see them.

There should probably be two buttons: one to deactivate the NSFW filter for only one page view, and another one to deactivate it permanently (by setting a cookie).

longj724:
Would something like what is shown in the video work? (this is just a proof of concept) If the user opts in to always seeing unsettling profiles then they will not be shown this prompt whenever they visit an unsettling profile.

https://user-images.githubusercontent.com/44875087/142796717-b05bca5f-217f-4628-aa2d-74792239cff6.mov

Changaco:
It's a good start.

- [ ] The profile should be completely hidden as long as the visitor hasn't opted-in to see it, including and especially the avatar.
- [ ] I think the warning needs to be more detailed and complete. Something like “This page is marked as containing potentially upsetting or embarrassing content. Viewing it is particularly unrecommended if someone else can see your screen or if you're a minor.”
- [ ] The first button should be labelled “View this page” instead of “View unsettling profile”.
- [ ] The text of the second button should end with “on this computer” or “in this browser” to clarify the scope of the opt-in, and it should use the word “pages” instead of “profiles” to be more general.
- [ ] The two buttons should probably have the `btn-warning` class instead of `btn-primary`.
[...]

PR #2098, opened 2021-12-09
Setting Profiles as Unsettling

longj724:
For issue #1965. Implements functionality that allows users to mark their profile as unsettling. These profiles are then blurred/hidden when viewed by other users. Users can only see the content of unsettling profiles after being warned the profile contains unsettling content and confirming that they still want to view the profile.


www/search.spt
```diff
@@ -154,12 +158,19 @@ if query:
...
+                    </a>
+                        {{ _(
+                        "{username} ",
+                        username=('<a href="/{0}/">{0}</a>'|safe).format(owner.username),
+                    ) }}
+                </div>
                 {{ _(
-                    "{username} has a repository named {repo_name} in their {platform} account",
+                    "has a repository named {repo_name} in their {platform} account",
                     username=('<a href="/{0}/">{0}</a>'|safe).format(owner.username),
```
Changaco, line 174: If the `username` variable isn't used in the string anymore, then it should be removed from the arguments passed to the translation function. However, the larger problem here is that this change assumes that the username is at the beginning of the sentence in every language. A cleaner solution would be to generate the possibly-blurred avatar+username box with a macro or [block assignment](https://jinja.palletsprojects.com/en/3.0.x/templates/#block-assignments), and pass it to the translation function in the `username` argument.

Changaco (changes requested):
- [ ] JavaScript isn't needed (and therefore shouldn't be used) to implement opting-in to view unsettling profiles.
- [ ] The CSS classes used to apply the blur effect can be generalized, i.e. `blur-3px` and `blur-4px` instead of `panel-blur`, `repo-search-blur` and `mini-user-blur`.
- [ ] The `base.html` template isn't the right place to implement hiding unsettling profiles. `profile-base.html` is more appropriate.

Changaco:
- [ ] redirect
- [ ] rebase

Changaco:
- [ ] The `privacy.spt` file can't only be renamed to `visibility.spt`, it has to keep existing to handle submissions of the old form. It can be a simple 307 redirect to the new URL, but in that case the `visibility.spt` file has to handle receiving a `privacy` parameter in the request body.
- [ ] This branch needs to be rebased on `master`.

PR #2104, opened 2021-12-14
improvement: sort accepted currencies on the "Donate" page alphabetically

WofWca:
It makes it easier to find the one you want if the list is big

I did not test this change in any way.

### DIFF
--- a/templates/macros/your-tip.html
+++ b/templates/macros/your-tip.html
@@ -82,10 +82,12 @@
             <form action="" method="GET" class="form-inline">
                 <select class="form-control" name="currency">
                 % set paypal = tippee_p.payment_providers.__and__(2)
-                % for c in accepted_currencies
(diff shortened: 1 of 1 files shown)


WofWca:
Thanks for checking this out, I'll come back to this later.
Looks like we need to move the logic to a `.py` file.

templates/macros/your-tip.html
```diff
@@ -82,10 +82,12 @@
             <form action="" method="GET" class="form-inline">
                 <select class="form-control" name="currency">
                 % set paypal = tippee_p.payment_providers.__and__(2)
-                % for c in accepted_currencies
+                % set get_title = lambda c: locale.title(locale.currencies.get(c, c))
+                % set accepted_currencies_sorted = sorted(accepted_currencies, key=get_title)
```
WofWca, line 86: It'd be cool to improve performance by adding caching later.

Changaco:
I included this in #2156.

PR #2120, opened 2022-01-27
Fixes for recent strings

comradekingu:


Changaco:
In my opinion none of the proposed changes is necessary, and most of them would actually make things worse.

liberapay/constants.py
```diff
@@ -344,8 +344,8 @@ def __missing__(self, currency):
 )
 
 PRIVACY_FIELDS = OrderedDict([
-    ('hide_giving', (_("Do not publish the amounts of money I send."), False)),
-    ('hide_receiving', (_("Do not publish the amounts of money I receive."), False)),
+    ('hide_giving', (_("Do not publish the amount of money I send."), False)),
```
jorgesumle, line 347: https://english.stackexchange.com/questions/254339/when-to-use-amount-vs-amounts

jorgesumle:
Using dots instead of commas in some places may be better, I would also write that way. But I am not sure that using commas there is wrong. I wouldn't change anything unless I'm sure it's wrong as it is.

liberapay/constants.py
comradekingu: Yes. This is also wrong in "Source code of the recurrent donations platform Liberapay"

templates/macros/your-tip.html
```diff
@@ -236,17 +236,17 @@ <h5 class="list-group-item-heading">{{ _("Manual renewal") }}</h5>
     % endif
     % if patron_visibilities == 1
     <p class="text-info">{{ glyphicon('info-sign') }} {{ _(
-        "{username} has chosen not to see who their patrons are, so your donation will be secret.",
```
comradekingu, line 239: This is a known singular, where "their" would denote plural.

templates/macros/your-tip.html
```diff
@@ -236,17 +236,17 @@ <h5 class="list-group-item-heading">{{ _("Manual renewal") }}</h5>
...
+        "{username} has chosen not to see own patrons, so your donation will be secret.",
         username=tippee_name
     ) }}</p>
     % elif patron_visibilities == 2
     <p class="text-info">{{ glyphicon('info-sign') }} {{ _(
-        "This donation won't be secret, you will appear in {username}'s private list of patrons.",
```
comradekingu, line 244: Sentence splicing.

templates/macros/your-tip.html
```diff
@@ -300,7 +300,7 @@ <h5 class="list-group-item-heading">{{ _("Public donation") }}</h5>
     </ul>
     % else
     <p class="text-muted">{{ glyphicon('info-sign') }} {{ _(
-        "{username} hasn't yet specified whether they want to see who their patrons are, "
```
comradekingu, line 303: Singular known they/their is a poor choice of words.

www/%username/giving/index.html.spt
```diff
@@ -227,8 +227,8 @@ next_payday = compute_next_payday_date()
 % elif show_incomplete_profile_warning
     <div class="text-info">
         {{ glyphicon('info-sign') }} {{ _(
-            "Since you've chosen to make a public donation, we recommend that you "
```
comradekingu, line 230: Run-On sentence

www/%username/giving/pay/stripe/%payin_id.spt
```diff
@@ -311,7 +311,7 @@ title = _("Funding your donations")
         ) }}</p>
         % if not route.has_been_charged_successfully()
         <p class="alert alert-warning">{{ _(
-            "Your bank can reject this payment. We recommend sending a copy of "
```
comradekingu, line 314: "We" is invoked without merit. It adds complexity.

www/%username/patrons/index.spt
```diff
@@ -76,10 +76,10 @@ subhead = _("Patrons")
 % if patron_visibilities > 1
 <h3>{{ _("Data export") }}</h3>
 <p><a href="{{ participant.path('patrons/export.csv?scope=active') }}">{{ fontawesome('download') }} {{
-    _("Download the list of currently active patrons")
```
comradekingu, line 79: "currently" is redundant. Am I missing something here?

www/%username/patrons/index.spt
```diff
@@ -76,10 +76,10 @@ subhead = _("Patrons")
...
+    _("Download the list of active patrons")
 }}</a></p>
 <p><a href="{{ participant.path('patrons/export.csv?scope=all') }}">{{ fontawesome('download') }} {{
-    _("Download the record of all patrons in the last ten years")
```
comradekingu, line 82: "In" is superfluous. It also makes it more of a garden-path sentence.

www/%username/patrons/index.spt
```diff
@@ -46,27 +46,27 @@ subhead = _("Patrons")
...
-            "then {link_start}click here to disable non-anonymous donations{link_end}.",
+            "then {link_start}click here to turn off non-anonymous donations{link_end}.",
             link_start='<button class="link" name="see_patrons" value="no">'|safe,
             link_end='</button>'|safe
         ) }}</p>
     % else
         <p class="text-muted">{{ _(
             "You've chosen not to see who your patrons are. If you change your mind, "
-            "then {link_start}click here to enable non-anonymous donations{link_end}.",
+            "then {link_start}click here allow non-anonymous donations{link_end}.",
```
comradekingu, line 69:
```suggestion
            "then {link_start}click here to allow non-anonymous donations{link_end}.",
```

www/%username/patrons/index.spt
```diff
@@ -46,27 +46,27 @@ subhead = _("Patrons")
...
             "You have opted-in to see who your patrons are. If you change your mind, "
-            "then {link_start}click here to disable non-anonymous donations{link_end}.",
+            "then {link_start}click here to turn off non-anonymous donations{link_end}.",
             link_start='<button class="link" name="see_patrons" value="no">'|safe,
             link_end='</button>'|safe
         ) }}</p>
     % else
         <p class="text-muted">{{ _(
             "You've chosen not to see who your patrons are. If you change your mind, "
-            "then {link_start}click here to enable non-anonymous donations{link_end}.",
```
comradekingu, line 69: It isn't clear what "enable" would do here.

www/%username/patrons/index.spt
```diff
@@ -46,27 +46,27 @@ subhead = _("Patrons")
     <input type="hidden" name="csrf_token" value="{{ csrf_token }}" />
     % if patron_visibilities == 0
         <p class="text-info">{{ glyphicon('info-sign') }} {{ _(
-            "Liberapay now supports non-anonymous donations, do you want to know "
+            "Liberapay now supports non-anonymous donations. Do you want to know "
             "who your patrons are?"
         ) }}</p>
         <p class="buttons">
             <button class="btn btn-primary" name="see_patrons" value="yes">{{ _(
-                "Enable non-anonymous donations"
```
comradekingu, line 54: Presumably this can't be disabled and on, so "enable" adds complexity.

comradekingu:
@Changaco @jorgesumle 
I added some rationale for the rest of the strings.
Most are repeat errors, so it is mostly explained once for each type.

templates/macros/your-tip.html
jorgesumle: Liberapay uses [singular they](https://en.wikipedia.org/wiki/Singular_they).

comradekingu: Hence why I made this PR to fix it.

jorgesumle:
@comradekingu, thanks for the explanation. I think that the only error is the comma splice (after reading [Comma Splice Examples and How to Fix Them](https://examples.yourdictionary.com/comma-splice-examples.html), I think that we need to fix that). The other corrections you proposed are mainly because you prefer simpler and more formal English, but I think that our more formal and verbose style is fine.

comradekingu:
@jorgesumle I proposed them because they are better, in ways I explained.
Simpler is always better when it means doing away with needless complexion.
I don't prefer less formal English. The examples of errors are not more formal, and some are unacceptable in any setting.

comradekingu:
@Changaco It is just a matter of avoiding issues. As an example "As Martin was a member of the board, they decided against the decision to allow the deconstruction." would be incomprehensible without interpreting as the language dictates.
Even if it didn't actually create these kinds of errors, conditions where a common pronoun is used for both sexes are exactly the same as everywhere else in Norway. There is no better control group or example in the world.

It is like changing the master branch naming to "main", only now the main lingua franca gets a lot harder to learn for people.

PR #2133, opened 2022-03-07
Remove Russia from the list of supported countries

Changaco:
[PayPal shuts down its services in Russia citing Ukraine aggression](https://www.reuters.com/business/paypal-shuts-down-its-services-russia-citing-ukraine-aggression-2022-03-05/). This means that creators in Russia can no longer use Liberapay to receive donations.

[Visa, Mastercard and American Express are also pulling out of Russia](https://www.reuters.com/business/american-express-suspends-operations-russia-belarus-2022-03-06/). This means that donors in Russia may currently be unable to fund their donations. However, Russian banks are working on switching to China's UnionPay, which we do support through Stripe.


CyberTailor:
UnionPay cards are not common in Russia and Stripe doesn't support Mir cards.

Can some other payment method (like Yandex) be added to Liberapay?

Changaco:
No, and because of the invasion I wouldn't do it even if it was possible.

CyberTailor:
> and because of the invasion I wouldn't do it even if it was possible

Sanctions against Internet users actually [harm](https://roskomsvoboda.org/uploads/documents/statement_of_roskomsvoboda_team.pdf) anti-war movement (for example, my local group can't use LP to receive donations for making posters/stickers).

Aspie96:
> No, and because of the invasion I wouldn't do it even if it was possible.

I understand in this case it isn't possible, but the attitude in the second sentence will not help us.

How many Russian oligarchs do you think rely on Liberapay? That's exact: none.

We should all condemn Putin's crimes in Ukraine, but our actions should be directed to people in power in Russia, not targeted against the general population that lives under an oppressive government.

We don't want the Russian people to be our enemy. Putin should be the common enemy.

ZelphirKaltstahl:
> We don't want the Russian people to be our enemy. Putin should be the common enemy.

While I agree on that, it has to be said, that far too many Russians kept silent in all those years Putin has been in power. And while it is certainly difficult to point exactly at the point in time, when they should have gone to the streets and protest, many are at least a tiny bit guilty of letting it come as far as it has come. Some have raised their voices, even in this hour and have paid a price for it, paying also for all the ones who kept silent again. Those who have raised their voices against the war, knowing well what could happen to them, are heros and have my admiration.

> We should all condemn Putin's crimes in Ukraine, but our actions should be directed to people in power in Russia, not targeted against the general population that lives under an oppressive government.

I would like there to be a way, in which the world can hit Putin and the ones supporting him exclusively, but unfortunately that is not easily possible. As long as people are not feeling anything amiss in their daily lives, they remain too complacent to do anything and Putin and his gang continue to rule.

jorgesumle:
When Russian banks support UnionPay, Russia can be added again.

Many people don't like Putin, but also many people don't like [NATO and their wars](https://en.wikipedia.org/wiki/List_of_NATO_operations), and this one looks a proxy war between NATO and Russia. There are many ongoing wars and some countries that Liberapay supports take part on them or send weapons. @Aspie96 pointed out that people don't use Liberapay to finance war, and I agree that we shouldn't block those people just because the government where they live is bad.

Changaco:
> When Russian banks support UnionPay, Russia can be added again.

No. This pull request removed Russia from the list of countries supported by PayPal, so it will only be reverted if PayPal becomes once again available to creators in Russia.

Issue #2134, opened 2022-03-08
Extraneous currency signs input by users aren't ignored


Changaco:
When a donor inputs `$10.00` or `10,00€` in a form, the amount is considered invalid and an error message is returned. It would be better to ignore the currency sign if it matches the expected currency.

biskwikman:
@Changaco I'm interested in doing this. Could you point me to the relevant file that deals with the donation field?

Changaco:
@biskwikman The `parse_money_amount` method in `liberapay/i18n/base.py`.

biskwikman:
@Changaco I just made a PR (#2150 ) which I seems to fix this issue.

PR #2150, opened 2022-05-30
Stop rejecting user-input money amounts which contain matching currency signs

biskwikman:
…t result in an error

I added the `get_currency_symbol` function from babel.

This PR is for issue #2134

### DIFF
--- a/liberapay/i18n/base.py
+++ b/liberapay/i18n/base.py
@@ -274,7 +274,7 @@ class Locale(babel.core.Locale):
         decimal_symbol = self.number_symbols['decimal']
         # If first character is relavent currency symbol, remove it and pass only numeric string
         # InvalidNumber will catch error if using the wrong currency symbol
-        if string[0] == get_currency_symbol(currency):
+        if string.startswith(self.currency_symbols.get(currency, currency)):
             string = string [1:]
         try:
             decimal = Decimal(
--- a/liberapay/i18n/base.py
+++ b/liberapay/i18n/base.py
@@ -7,7 +7,7 @@ from unicodedata import combining, normalize
 import babel.core
 from babel.dates import format_date, format_datetime, format_time, format_timedelta
 from babel.messages.pofile import Catalog
-from babel.numbers import parse_pattern
+from babel.numbers import parse_pattern, get_currency_symbol
 from cached_property import cached_property
 from markupsafe import Markup
 import opencc
@@ -272,6 +272,10 @@ class Locale(babel.core.Locale):
     def parse_money_amount(self, string, currency, maximum=D_MAX):
         group_symbol = self.number_symbols['group']
         decimal_symbol = self.number_symbols['decimal']
+        # If first character is relavent currency symbol, remove it and pass only numeric string
+        # InvalidNumber will catch error if using the wrong currency symbol
+        if string[0] == get_currency_symbol(currency):
+            string = string [1:]
         try:
             decimal = Decimal(
                 string.replace(group_symbol, '').replace(decimal_symbol, '.')
--- a/liberapay/i18n/base.py
+++ b/liberapay/i18n/base.py
@@ -7,7 +7,7 @@ from unicodedata import combining, normalize
 import babel.core
 from babel.dates import format_date, format_datetime, format_time, format_timedelta
 from babel.messages.pofile import Catalog
-from babel.numbers import parse_pattern, get_currency_symbol
+from babel.numbers import parse_pattern
 from cached_property import cached_property
 from markupsafe import Markup
 import opencc
--- a/liberapay/i18n/base.py
+++ b/liberapay/i18n/base.py
@@ -275,7 +275,7 @@ class Locale(babel.core.Locale):
         # If first character is relavent currency symbol, remove it and pass only numeric string
         # InvalidNumber will catch error if using the wrong currency symbol
         if string.startswith(self.currency_symbols.get(currency, currency)):
-            string = string [1:]
+            string = string[1:]
         try:
             decimal = Decimal(
                 string.replace(group_symbol, '').replace(decimal_symbol, '.')
--- a/liberapay/i18n/base.py
+++ b/liberapay/i18n/base.py
@@ -287,6 +287,8 @@ class Locale(babel.core.Locale):
         # then an `InvalidNumber` exception is raised.
         try:
             decimal = Decimal(
+                ''.join(string.split()).replace(decimal_symbol, '.')
+                if group_symbol.isspace() else
                 string.replace(group_symbol, '').replace(decimal_symbol, '.')
             )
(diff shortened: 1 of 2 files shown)


biskwikman:
oh, apologies for the whitespace that the test caught. I can resubmit if you want.

liberapay/i18n/base.py
```diff
@@ -272,6 +272,10 @@ def format_timedelta(self, o, **kw):
     def parse_money_amount(self, string, currency, maximum=D_MAX):
         group_symbol = self.number_symbols['group']
         decimal_symbol = self.number_symbols['decimal']
+        # If first character is relavent currency symbol, remove it and pass only numeric string
+        # InvalidNumber will catch error if using the wrong currency symbol
+        if string[0] == get_currency_symbol(currency):
```
Changaco, line 277: This is incorrect and inefficient in several ways:

1. `string[0] == ...` raises an `IndexError` if the string is empty, and creates a new string otherwise.
2. `get_currency_symbol(currency)` boils down to `Locale.parse(LC_NUMERIC).currency_symbols.get(currency, currency)`. The big problem with that is it uses the server's system locale, not the website user's locale. The minor problem is that it unnecessarily creates a transient `Locale` object. This small inefficiency is why Liberapay doesn't use those Babel functions anymore.

The proper and more efficient way do this is:

```suggestion
        if string.startswith(self.currency_symbols.get(currency, currency)):
```

liberapay/i18n/base.py
```diff
@@ -272,6 +272,10 @@ def format_timedelta(self, o, **kw):
...
+            string = string [1:]
```
Changaco, line 278:
```suggestion
            string = string[1:]
```

Changaco (changes requested):

liberapay/i18n/base.py
biskwikman: ah, good point about the empty string.

And good to know about the Babel functions.

liberapay/i18n/base.py
```diff
@@ -272,6 +272,10 @@ def format_timedelta(self, o, **kw):
     def parse_money_amount(self, string, currency, maximum=D_MAX):
         group_symbol = self.number_symbols['group']
         decimal_symbol = self.number_symbols['decimal']
+        # If first character is relavent currency symbol, remove it and pass only numeric string
+        # InvalidNumber will catch error if using the wrong currency symbol
+        if string.startswith(self.currency_symbols.get(currency, currency)):
+            string = string[1:]
```
Changaco, line 278: I didn't notice before, but this is incorrect. Firstly because the currency symbol isn't always a single character. Secondly because the previous line falls back to the three-letter ISO currency code when there is no symbol.

I also see two needed improvements:

1. We should also look for the currency symbol at the end of the string, because in most languages it's standard to put the currency symbol after the number, not before.
2. Strip the string, both before looking for the currency symbol and after it has been removed, because people put spaces everywhere, sometimes without realizing it.

Changaco (changes requested):

liberapay/i18n/base.py
biskwikman: Ok. I'll start working on that. By the way, how do I see what going on in the Python backend? For instance if I wanted the print the value of `string` at some point? I'm not experienced in Python on the web. I'm also pretty new to makefiles. Does `make run` start some kind of local Python server? Or is it something else?

Changaco: When you do a `make run`, two commands are executed, which you can see in the output:

1. The first one applies database migrations: `PYTHONPATH=. env/bin/python cli/run.py -e defaults.env,local.env env/bin/python liberapay/models/__init__.py`
2. The second one runs `app.py` (which launches the [Gunicorn](https://gunicorn.org/) HTTP server): `PATH=env/bin:$PATH env/bin/python cli/run.py -e defaults.env,local.env env/bin/python app.py`

Both commands first call `cli/run.py` to load environment variables from the `defaults.env` and `local.env` files.

If you add a `print(string)` line in the code, then the value will show up in your terminal. If the code is run through Gunicorn, then the printed value will be followed by at least one HTTP request log line, so don't expect it to be the last line of the output. If the code is run through pytest, then the printed value will be in the `Captured stdout call` section of the output for each failing test that encounters your print call.

biskwikman: This is what I've come up with. I've tested it with currency symbols of different lengths, with the symbol in the front and back, and with spaces between the numerical amount and the symbol, and at either end of the whole string.

```python
# If string begins or ends with relavent currency symbol, remove it and pass only numeric string
# InvalidNumber will catch error if using the wrong currency symbol
        string = string.strip()
        currency_length = len(self.currency_symbols.get(currency, currency))
        if string.startswith(self.currency_symbols.get(currency, currency)):
            string = string[currency_length:]
        elif string.endswith(self.currency_symbols.get(currency, currency)):
            string = string[0:-currency_length]
        string = string.strip()
```

Changaco (changes requested):
Three things still bother me:

1. For currencies which have symbols, the three-letter currency code isn't removed if present. I think it would be better to support the ISO codes either universally or not at all.
2. The currency symbol is looked up three times. There's no reason to do it more than once.
3. The comment contains a typo ('relavent').

biskwikman:
I added changes which address points 2 and 3. I'm not quite sure what you mean by 1. But would it be resolved by removing the second `currency` from `currency_symbol = self.currency_symbols.get(currency, currency)`?

Changaco:
@biskwikman We forgot to add a test to prevent regressions. Do you want to try to write it? The appropriate file is `tests/py/test_i18n.py`.

biskwikman:
@Changaco Sure. Would this be sufficient?
 
 ```
def test_parse_money_amount_rejects_incorrect_currency_symbols(self):
    with self.assertRaises(InvalidNumber):
        LOCALE_EN.parse_money_amount("$100", 'EUR')
```

**Comment by Changaco** [comment]
@biskwikman That test is good but it wouldn't detect regressions since it only checks a behaviour that isn't new. We need to test that "€100", "100€" and " 100 € " are all accepted.

**Comment by biskwikman** [comment]
@Changaco 
ok. I'm a little unsure but this seems to work the way it should when I run the tests.

```python
def test_parse_money_amount_rejects_incorrect_currency_symbols(self):
    with self.assertRaises(InvalidNumber):
         LOCALE_EN.parse_money_amount("$100", 'EUR')
     assert LOCALE_EN.parse_money_amount("€100", 'EUR')
     assert LOCALE_EN.parse_money_amount("€ 100", 'EUR')
     assert LOCALE_EN.parse_money_amount("100€", 'EUR')
     assert LOCALE_EN.parse_money_amount("100 €", 'EUR')
```

Changaco:
Squashed and merged manually.

PR #2158, opened 2022-06-15
Add Croatia, Gibraltar and Liechtenstein to the list of countries supported by Stripe

Changaco:
Liberapay is now fully available in the entire European Union, as Croatia was the last unsupported EU country.


Changaco:
Iceland is now the last missing [EEA](https://en.wikipedia.org/wiki/European_Economic_Area) and [EFTA](https://en.wikipedia.org/wiki/European_Free_Trade_Association) country.

PR #2164, opened 2022-07-14
added public_name search function

### LINKED ISSUE #2082: Searching for the last name does not find an account
I just used the Liberapay search for "Owens". The profile I was looking for is this:

https://en.liberapay.com/doctormo/

and the name "Owens" is known to Liberapay as can be seen in the profile. Unfortunately, the Liberapay search only found unrelated accounts.

biskwikman:
closes #2082 

I basically mimicked the structure of the code for usernames, but pointed the new query to the `public_name` column of the `participants` table, it seems to work fine.


Changaco (changes requested):
I think it's probably better to have a single query and unified results section for both usernames and public names.

Also, a new [trigram index](https://www.postgresql.org/docs/current/pgtrgm.html) is needed on the `public_name` column to speed up the search:

```sql
CREATE INDEX public_name_trgm_idx ON participants
    USING GIN (lower(public_name) gin_trgm_ops)
    WHERE status = 'active'
      AND public_name IS NOT null;
```

biskwikman:
Should we include the username and public name under the avatar on all search results from now on? I think it would help with any confusion if an account has a similar username and public name.

Changaco:
Yes, both names should be displayed for all results.

biskwikman:
I've made it so the username and public name are returned in a single query, and both are displayed. I also added the index.

I'm not sure what word to use instead of "username" now. For example in the phrasing "Found a matching username". I was thinking of just using "user", but maybe "individual" would be more consistent.

www/search.spt
```diff
@@ -15,12 +15,23 @@ if query:
...
+                 , id
               FROM participants
-             WHERE status = 'active'
-               AND NOT username like '~%%'
-               AND lower(username) %% lower(%(q)s)
+             WHERE lower(username) %% lower(%(q)s)
+                OR lower(public_name) %% lower(%(q)s)
+               AND status = 'active'
                AND hide_from_search = 0
+               AND NOT username like '~%%'
```
Changaco, line 34: This is incorrect due to the order of precedence of the boolean operators. `… OR … AND …` is equivalent to `… OR ( … AND … )`, but here you wanted `( … OR … ) AND …`.

Also, you forgot to add `public_name IS NOT NULL`, so the query isn't using the new index.

```suggestion
             WHERE status = 'active'
               AND hide_from_search = 0
               AND ( NOT username like '~%%' AND lower(username) %% lower(%(q)s)
                     OR
                     public_name IS NOT NULL AND lower(public_name) %% lower(%(q)s)
                   )
```

www/search.spt
```diff
@@ -15,12 +15,23 @@ if query:
...
+            SELECT username
+                 , public_name
+                 , avatar_url
+                 , CASE 
+                        WHEN public_name IS NULL 
+                             THEN similarity(username, %(q)s)
+                        WHEN similarity(username, %(q)s) >= similarity(public_name, %(q)s) 
+                             THEN similarity(username, %(q)s)
+                        ELSE similarity(public_name, %(q)s)
+                    END AS rank
```
Changaco, line 27:
```suggestion
                 , greatest(
                       similarity(username, %(q)s),
                       similarity(public_name, %(q)s)
                   ) AS rank
```

Changaco (changes requested):

biskwikman:
thanks for the suggestions. I just wanted to check, the tests will have to be updated for this, right? It looks like two are failing now.

Changaco:
The tests aren't failing with the latest commit. I think GitHub is still showing earlier test results because there's currently a merge conflict (because #2177 hasn't been deployed yet, because I've been waiting for new machine and human translations (https://github.com/liberapay/salon/issues/520)).

Changaco:
Sorry I let this linger for two years. It's now merged (#2401) and will be deployed this week.

PR #2172, opened 2022-08-06
optimized icon-v2_white-on-yellow.svg

RiedleroD:
the original was a bit sloppy (e.g. the yellow rectangle wasn't perfectly symmetric), so it's not pixel-perfectly the same, but close enough for nobody to notice the difference without a direct comparison.

What I changed:
- resolved grouping
- replaced background with <rect> (was a path before)
- better code formatting for readability
- resolved matrix transform
- optimized paths
  - converted curves to short curves where possible
  - moved starting points to take advantage of z
  - fixed rounding errors)
  - generally more compact notation

Unlike with my usual optimized SVGs, there were no elliptical arcs, so there shouldn't be any problems with bad SVG renderers like rsvg and QtSVG.


RiedleroD:
as to why I did this: I had to use the liberapay icon on my website anyway, and I don't want bloat like unoptimized SVGs there, so I thought why not contribute while I'm at it.

PR #2174, opened 2022-08-15
Fix nav overflow

vrifox:
For to long Strings (e.g. translated) it's possible to overflow (at least) the left navigation. This proposed Pull request should fix that.

| before | after |
| -------- | ------ |
| grafik | grafik |


Changaco:
Thanks.

PR #2175, opened 2022-08-16
Fix searchbar width

biskwikman:
Inspired by #2174, I tried to fix the issue where the search input element's width was too small to show the entire placeholder. I made the containing div fit the input's width.


biskwikman:
Oops, I messed up this pr. I'm going to close and retry.

PR #2176, opened 2022-08-16
make width of search div that of search input

biskwikman:
Inspired by #2174, I tried to fix the issue where the search input element's width was too small to show the entire placeholder. I made the containing div fit the input's width.

I messed up the previous PR for this. I closed that and this is the correct one.

### DIFF
--- a/style/base/base.scss
+++ b/style/base/base.scss
@@ -357,6 +357,10 @@ ul.memberships {
     }
 }
 
+div[role=search] {
+    width: max-content;
+}
+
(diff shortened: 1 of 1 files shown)


Changaco:
The search `<input>` is supposed to be 250 pixels wide, but the CSS rule for that is overshadowed by one of Bootstrap's rules. Changing our rule to set a minimum width avoids this overshadowing, but a future upgrade of Bootstrap could break things again.

PR #2178, opened 2022-08-28
Various translation-related improvements

Changaco:


jorgesumle:
I see the machine-translation warning in every page, for both Esperanto and Spanish. However, they aren't fuzzy on Weblate.

Changaco:
That's because I hadn't deployed the changes of the last two weeks.

PR #2183, opened 2022-10-09
Fix email address leak

Changaco:
The short-lived tool to migrate one's account from Gratipay contained a design flaw that could be used to obtain the primary email address of any Liberapay account. This branch fixes that leak, and removes the data importer entirely.


Changaco:
I've found no evidence of this vulnerability being exploited in the logs with have, which cover the last 2 years.

I'm not sure if we should notify the data protection authority. What do you think @jorgesumle?

Changaco:
Article 33 of the [GDPR](https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=CELEX%3A32016R0679) states that a personal data breach must be notified “unless [it] is unlikely to result in a risk to the rights and freedoms of natural persons”. Considering that this vulnerability doesn't seem to have been exploited, I think it is indeed unlikely to cause problems for the people who could have been affected.

PR #2189, opened 2022-11-03
Various

Changaco:


Changaco:
This branch includes visible changes to the payment pages. Here's an example with a donor who has three donations awaiting payment, one of them in a currency that isn't accepted by the creator anymore:

Screenshot of the modified payments page

PR #2199, opened 2022-12-17
improvement: add icon with yellow stroke

WofWca:
| Before | After |
|-|-|
| old on white | new on white
| old on black | new on black |

Closer look:

image

TODO:

* [ ] Replace the favicon.
* [ ] Generate PNGs.
* [ ] Search for occurrences of "icon-v2_black" in the codebase and replace where appropriate.
* [ ] Consider white stroke instead of yellow.

I'll do these if you like the change.


trebmuh:
nice looking

hazem454:
Can call my Whatssapp 0096567088116

Changaco:
It doesn't look quite right to me. Maybe if the outline was thinner or had an alpha gradient?

Is this change only for the favicon, or do you have another reason?

WofWca:
For the favicon in the first place. Maybe there are other places it can be useful in.

WofWca:
> Maybe if the outline was thinner or had an alpha gradient?

I have picked width that looked the best to me at 16px width. You can play around with it in the dev tools as well.
For the gradient - currently I'm not sure how to achieve it.

WofWca:
Hey, I also just learned that you can set different favicons depending on color scheme, with the [`media` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-media)

Changaco:
@WofWca I assume you're talking about [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme). It doesn't seem to work on favicons in Firefox.

Issue #1656, opened 2020-02-07
Cron jobs


Changaco:
#1652 has added cron jobs that perform critical work (they're responsible for the execution of scheduled payments), so we need to improve the reliability and transparency of our cron system.

PR #2224, opened 2023-02-23
Fix corner case in payments to teams

Changaco:
On February 1st, a member of [the Mobian team](https://liberapay.com/mobian/), @a-wai, changed his take from `auto` to €250. Since [manual takes have priority](https://liberapay.com/about/teams), and the team's income is less than €250, this change has redirected all the team's income to him. Unfortunately, the fact that his Stripe account is disabled, combined with the bug fixed by this commit, prevented executing any new grouped payment which included the Mobian team as one of the beneficiaries. This snowballed into unrelated automatic payments not being executed on their scheduled dates (<https://github.com/liberapay/salon/issues/542#issuecomment-1440781746>), because it crashed the daily “cron” job. Finally, this entire problem went undetected for two weeks due to several shortcomings:

- for an as-yet unknown reason, the crashes of the daily cron job appear not to have been sent to, or recorded by, Sentry;
  - this might be due to the fact that we're not using the latest version of the Sentry Python library, so it's possible that a policy of regular updates of dependencies would have avoided this
- the limited amount of information displayed in the `/admin/schedules` page did not enable me to realize that some automatic payments were not being executed
- I have failed to fix in a timely manner a weakness which I had identified (#1656)

I plan to address these shortcomings in subsequent pull requests.

### DIFF
--- a/liberapay/models/_mixin_team.py
+++ b/liberapay/models/_mixin_team.py
@@ -202,17 +202,11 @@ class MixinTeam:
     def get_current_takes_for_payment(self, currency, weekly_amount):
         """
         Return the list of current takes with the extra information that the
-        `liberapay.payin.common.resolve_take_amounts` function needs to compute
+        `liberapay.payin.common.resolve_team_donation` function needs to compute
         transfer amounts.
         """
         takes = self.db.all("""
-            SELECT t.member
-                 , t.ctime
-                 , convert(t.amount, %(currency)s) AS amount
-                 , convert(
-                       coalesce_currency_amount(t.paid_in_advance, t.amount::currency),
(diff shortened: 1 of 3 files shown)


a-wai:
Ouch, I wouldn't have expected this action to cause such an issue! Sorry for not addressing the stripe problem earlier (which is now done, fwiw) and many thanks for your work on liberapay!

Changaco:
I've figured out why the failure of the daily background task wasn't recorded. The type of exception which was raised is supposed to be ignored when in the context of processing a user request, but the error handler doesn't actually check that a request is being processed. The good news is that this is easy to fix and not at all Sentry's fault.

PR #2302, opened 2023-12-22
Enable creators to choose which countries donations are allowed to come from

### LINKED ISSUE #1102: Nationalities of donors
Some organizations cannot accept donations from foreign nationals, thus they are currently unable to use Liberapay.

Changaco:
Closes #1102 (even though this implementation is actually based on territoriality, not nationality).


Changaco:
Screenshot of the new page where a user can select the countries they want to accept donations from. It's a long multi-column list of checkboxes, one per territory.

Issue #202, opened 2016-03-04
Allow users to set a custom avatar URL


Changaco:
This requires a proxy that fetches the URL, checks that it's a valid image, and resizes it if necessary. Either we find a free and open service that does that, or we set one up ourselves.

Changaco:
Here's a related blog post: [My fight against CDN libraries - Pepper&Carrot](http://peppercarrot.com/en/article390/my-fight-against-cdn-libraries).

renyuneyun:
Hi, are there any news on the plan of this?
I struggled for a while and now realised this is not implemented :crying_cat_face:

PR #2334, opened 2024-04-04
Properly implement the libravatar federation API

### LINKED ISSUE #1367: Federated libravatar should be supported
The key feature of libravatar is federation. Please add support for it. The necessary steps are roughly described in https://wiki.libravatar.org/api/ and the available libraries are listed in https://wiki.libravatar.org/libraries/.

LukeShu:
I have no strong opinions about whether WebFinger should be supported instead of Libravatar (as suggested in #1043), but as long as there's a thing in Liberapay that says "libravatar", it should do what the [libravatar API docs](https://wiki.libravatar.org/api/) say. It is surprising and confusing that Liberapay will tell users "We were unable to get an avatar for you from libravatar" while https://www.libravatar.org/tools/check/ is telling them that their avatar is A-OK.

Unlike the old https://github.com/liberapay/liberapay.com/pull/1043, this implements the libravatar API in Liberapay, instead of using the unmaintained (since 2016) pyLibravatar library (which in turn uses the py3dns library, instead of the Dnspython library that Liberapay prefers).

I have done minimal testing of this locally on my laptop.

Closes #1367

### DIFF
--- a/liberapay/models/participant.py
+++ b/liberapay/models/participant.py
@@ -6,6 +6,7 @@ from email.utils import formataddr
 from hashlib import pbkdf2_hmac, md5, sha1
 from operator import attrgetter, itemgetter
 from os import urandom
+from random import randint
 from threading import Lock
 from time import sleep
 from types import SimpleNamespace
@@ -16,6 +17,7 @@ import uuid
 import aspen_jinja2_renderer
 from cached_property import cached_property
 from dateutil.parser import parse as parse_date
+from dns.resolver import Cache as DNSCache, Resolver as DNSResolver
 from html2text import html2text
 from markupsafe import escape as htmlescape
 from pando import json, Response
@@ -97,6 +99,10 @@ TEN_YEARS = timedelta(days=3652)
 
 
 email_lock = Lock()
+
+DNS = DNSResolver()
+DNS.lifetime = 1.0  # 1 second timeout, per https://github.com/liberapay/liberapay.com/pull/1043#issuecomment-377891723
+DNS.cache = DNSCache()
 
 
 class Participant(Model, MixinTeam):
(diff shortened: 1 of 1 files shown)


liberapay/models/participant.py
```diff
@@ -2176,8 +2182,49 @@ def update_avatar(self, src=None, cursor=None, avatar_email=None, check=True):
...
+                top_priority = min(rec.priority for rec in srv_records)
+                srv_records = [rec for rec in srv_records if rec.priority == top_priority]
+                # Of those, choose randomly based on their relative `.weight`s.
+                weight_choice = randint(0, sum(rec.weight for rec in srv_records))
+                weight_sum = 0
+                for rec in srv_records:
+                    weight_sum += rec.weight
+                    if weight_sum >= weight_choice:
+                        choice_record = rec
+                        break
```
Changaco, line 2207: Is this algorithm standard? Can you add a link to where you found it?

LukeShu: Yes, this is the standard RFC2782 algorithm (described at the top of page 4 of the RFC). I'll amend the commit to add a comment.

Also, I just noticed a bug in my implementation with zero-weight records.

LukeShu:
v3:
- Add more comments
- add `srv_records.sort(key=lambda rec: rec.weight)  # ensure that .weight=0 recs are first in the list` per RFC 2782

```patch
diff --git a/liberapay/models/participant.py b/liberapay/models/participant.py
index 9995e0775..4d42b01db 100644
--- a/liberapay/models/participant.py
+++ b/liberapay/models/participant.py
@@ -2183,6 +2183,10 @@ class Participant(Model, MixinTeam):
             if not email:
                 return
             # https://wiki.libravatar.org/api/
+            #
+            # We only use the first SRV record that we choose; if there is an
+            # error talking to that server, we give up, instead of retrying with
+            # another record.  pyLibravatar does the same.
             normalized_email = email.strip().lower()
             try:
                 # Look up the SRV record to use.
@@ -2194,10 +2198,20 @@ class Participant(Model, MixinTeam):
                     srv_records = DNS.query('_avatars._tcp.'+email_domain, 'SRV')
                     scheme = 'http'
                 # Filter down to just the records with the "highest" `.priority`
-                # (lower number = higher priority).
+                # (lower number = higher priority); for the libravatar API tells us:
+                #
+                # > Libravatar clients MUST only consider servers listed in the
+                # > highest SRV priority.
                 top_priority = min(rec.priority for rec in srv_records)
                 srv_records = [rec for rec in srv_records if rec.priority == top_priority]
-                # Of those, choose randomly based on their relative `.weight`s.
+                # Of those, choose randomly based on their relative `.weight`s;
+                # for the libravatar API tells us:
+                #
+                # > They MUST honour relative weights.
+                #
+                # RFC 2782 (at the top of page 4) gives us this algorithm for
+                # randomly selecting a record based on the weights:
+                srv_records.sort(key=lambda rec: rec.weight)  # ensure that .weight=0 recs are first in the list
                 weight_choice = randint(0, sum(rec.weight for rec in srv_records))
                 weight_sum = 0
                 for rec in srv_records:
```

liberapay/models/participant.py
LukeShu: Done.

Changaco:
Privacy impact assessment: as it was already possible to inject arbitrary avatar URLs through the Mastodon integration, implementing the Libravatar federation protocol doesn't significantly increase the possibility for third parties to observe who is visiting the Liberapay website.

Security considerations: instructing browsers to download images from untrusted domains is a risk that should be mitigated by proxying (as initially envisaged in #202), but since this risk already exists and isn't being significantly increased by the implementation of the Libravatar federation protocol, it doesn't justify holding this pull request.

PR #2346, opened 2024-04-22
This commit introduces a new feature that checks if participants' inc…

cloud-leon:
Changes include: - Updated SQL schema to add 'weekly_income_goal' to the 'participants' table and create an 'income_notifications' table for tracking sent notifications. - Added methods in the  model:   
-  to perform the check against the set income goal.   
-  to calculate the total received income within a specific date range.   -  to determine if a notification was recently sent.  
-  -  to handle the sending of notifications and record them in the database.
-  - Implemented tests in a new file :   
- Test for the scenario where the income goal is met and a notification is sent.   
- Test for the scenario where the income goal is not met.   - Test to ensure no notification is sent if one was sent recently.  These additions enhance user engagement by actively notifying users about their income status in relation to their goals, improving the platform's support for financial tracking and user motivation.


Changaco:
There are so many mistakes in this patch that I'm not sure it's worth writing a detailed review.

PR #2383, opened 2024-06-06
Switch back to the old Sources API for SEPA Direct Debit

Changaco:
Reverses #2376, because it results in the postal address of the recipient of a direct debit showing up in the SEPA mandate, which is a privacy hazard.


ftservice (approved):

ftservice:
Good 2

PR #2525, opened 2025-01-13
Stop linking to X/Twitter

jorgesumle:
See https://github.com/liberapay/salon/issues/656


Changaco:
We should probably remove the Facebook links too.

PR #2540, opened 2025-04-01
Hold high risk charges for review

Changaco:
This branch changes the way Liberapay submits card charges to Stripe. It separates the authorization and capture steps. That enables checking the risk level that Stripe computes during the first step before initiating the second step. The goal is to reduce the number of fraudulent charges that go through, and consequently the number of refunds we have to do and the amount of money we lose on those.

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -213,6 +213,19 @@ Liberapay.forms.jsSubmit = function() {
                             window.location.href = window.location.href;
                             navigating = true;
                         }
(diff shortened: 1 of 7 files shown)


Changaco:
There were multiple flaws in this, and it caused some problems for users, but it also stopped 8 fraudulent charges on its first day.

Hotfixes: https://github.com/liberapay/liberapay.com/compare/784...789

PR #2553, opened 2025-04-25
UI - Low quality WebP/PNG icons are used instead of SVG for some platforms

h7p3ri0n:
SVG icons were not used for Mastodon, Gitlab and LinuxFr.org, even if they are present in the `/~/assets/platforms` folder. OpenStreetMap didn't have one, so I added it. These changes were tested on Firefox and Chromium and seems to resolve the conflict between old icons and SVG ones for these platforms.

ksnip_20250424-013913


Changaco:
The reason some icons are served as PNG or WebP is that the alternative formats are significantly less efficient for these specific images at these specific sizes, meaning that they result in significantly larger files and thus more network traffic, even after being optimized.

PR #2611, opened 2025-08-16
Fix German string

salim-b:


Changaco:
In the future, please submit translation changes [through Weblate](https://hosted.weblate.org/projects/liberapay/core/de/) instead.

PR #2641, opened 2025-10-23
Public team ledger

Changaco:
This branch creates a new public page showing how each donation to a team is split between its members. It also makes the data available in CSV and JSON formats.

One thing I'm not sure about is whether the donor's payment method should be revealed. Recipients already have access to that information, but this would be the first time it's disclosed publicly.

### DIFF
--- a/liberapay/renderers/csv_dump.py
+++ b/liberapay/renderers/csv_dump.py
@@ -11,9 +11,14 @@ class Renderer(renderers.Renderer):
         if not rows:
             return ''
         f = StringIO()
-        w = csv.writer(f)
         if hasattr(rows[0], '_fields'):
+            w = csv.writer(f)
             w.writerow(rows[0]._fields)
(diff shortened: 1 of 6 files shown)


Changaco:
Sample screenshot of the new public team ledger at 800px width

---

On small screens it doesn't look great, but it doesn't overflow.

Sample screenshot of the new public team ledger at 360px width

Changaco:
The reason to show the payment method would be the same reason to show anything else: transparency. The use case could be statistics. If Liberapay doesn't publish the method of each payment, then people can't do the statistics on their own, so ideally Liberapay should publish aggregated data on payment methods.

Changaco:
I'm leaning towards showing the payment method, and probably also whether the payment was automatic or manually triggered by the user.

Relevant fact: Open Collective publicly shows the method of each payment.

PR #2711, opened 2026-02-10
Upgrade Canada and the US to top tier

Changaco:
Stripe has effectively added those two countries to what was previously its European region. Liberapay is now able to make separate transfers to accounts in Canada and the US, thus enabling grouped payments to those countries and re-enabling SEPA Direct Debits to them. However, we remain unable to support grouped payments and SEPA Direct Debits to seven other territories (Australia, Hong Kong, Japan, Malaysia, Mexico, New Zealand, Singapore).

### DIFF
--- a/tests/py/fixtures/TestAvatars.yml
+++ b/tests/py/fixtures/TestAvatars.yml
@@ -8,48 +8,23 @@ interactions:
     body: {string: ''}
     headers:
       Access-Control-Allow-Origin: ['*']
-      Cache-Control: [max-age=300]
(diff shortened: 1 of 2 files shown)


Changaco:
This is blowing up in my face. I forgot to actually test a separate transfer to Canada, and it turns out Stripe's documentation is wrong about those transfers being allowed. Moreover, it hadn't occurred to me that making these changes would trigger a big wave of SEPA Direct Debits for donations that have been in limbo since that payment method was restricted to European recipients.

PR #2732, opened 2026-03-05
add automatic retries of PayPal API requests

### LINKED ISSUE #2731: Automatic retries of PayPal API requests
The `liberapay/payin/paypal.py` file should be modified so that API requests are retried automatically once or twice before giving up and raising an exception. This would prevent some payments from failing when a transient network or hardware failure occurs.

(The `stripe` module already retries requests automatically.)

ojsed:
Configure the requests.Session used for PayPal API calls with a urllib3 Retry adapter: 2 retries with exponential backoff on 5xx server errors. This matches the Stripe module's retry configuration (stripe.max_network_retries = 2) and prevents some payments from failing due to transient network or hardware issues.

Closes #2731

### DIFF
--- a/liberapay/payin/paypal.py
+++ b/liberapay/payin/paypal.py
@@ -4,7 +4,9 @@ import re
 from time import sleep
 
 import requests
+from requests.adapters import HTTPAdapter
 from pando.utils import utcnow
+from urllib3.util.retry import Retry
 
 from ..exceptions import (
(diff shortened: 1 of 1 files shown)


Changaco (changes requested):
The proposed change doesn't fully fix the problem. All requests that fail to return JSON data should be retried, regardless of their method and the response status code.

Issue #75, opened 2016-01-23
Improve translation workflow


Changaco:
Issues with Weblate:

- its locking [can be annoying](https://github.com/liberapay/salon/issues/2#issuecomment-173942861) (https://github.com/nijel/weblate/issues/245)
- I had to [hack my way around](https://github.com/liberapay/liberapay.com/pull/53) its lack of a proper review system (https://github.com/nijel/weblate/issues/867)

Changaco:
There's something else bothering me: why don't we have easy i18n of simple strings like button labels? Those translations have been done lots of times already in other projects, we shouldn't have to spend time on them.

Changaco:
A translator working while I was reviewing and merging translations has led to a git conflict. :-/

Changaco:
The worse part: the conflict was only in those fucking useless PO file headers.

EliotBerriot:
Regarding the merge conflict issue, maybe [this kind of setup](https://docs.weblate.org/en/latest/admin/continuous.html#avoiding-merge-conflicts) could offer a solution ?

Changaco:
It looks similar to what we do now, except we don't put a lock. We also don't use a script to merge PO headers (it's rare enough to be handled manually).

We can't automatically push new translations to master because we need to review them first (we use Git's builtin tools to do that since Weblate doesn't give us a way to do it).

Changaco:
[Weblate 2.18 introduces a review system and replaces the locking system with conflict detection](http://lists.cihar.com/pipermail/hosted-weblate-announce/2017-December/000000.html).

jorgesumle:
One thing that could be improved is [adding an explanation](https://docs.weblate.org/en/latest/admin/translating.html#additional-info-on-source-strings). That could be of helps in strings like [this](https://hosted.weblate.org/translate/liberapay/core/es/?checksum=9e2d30f1d2a456e3), where some translators translated the article title, which in my opinion is not the proper approach. Explanation could here say, "the article is only in English. Do not translate the title. If you want, add the language of the linked article as an clarification".

buhtz:
Maybe not optimal for your workflow but could be an inspiration. I am using a [self-written Python script](https://github.com/bit-team/backintime/blob/dev/update_language_files.py) (including [my own docu](https://github.com/bit-team/backintime/blob/dev/doc/maintain/2_localization.md)) to sync my Python sources with the Weblate instance at translate.coderberg.org. Additionally that script does do a lot of extra work.

[...]

PR #2782, opened 2026-06-20
2FA - OTP + WebAuthn

8by3:
I signed up to Liberapay about a year ago and only actually added it to my projects this week and was really surprised you don't support 2fa, its really table stakes in 2026! So I added it :-)

I've not written Python in anger for a good ten years so please forgive me if its not very idiomatic. To reduce new code written, I used what looked like respectable deps for the otp/webauthn itself.

One assumption made was to reuse SECRET_FERNET_KEYS to encrypt the 2fa secret, instead of requiring a new secret to be added.

The PR is rather large as it includes translations for the 38 supported languages, these were machine generated, so I can't vouch for their 100% accuracy but for simple strings used translations are pretty reliable.

I don't normally use github, I've joined the salon on matrix if you have any questions @s:8by3.net

Best,

sbr.


8by3:
@Changaco Hey, is there anything I can do to make this less effort?

Changaco:
@8by3 Yes. Split the changes into smaller commits. TOTP, Passkeys and recovery codes are three separate features and ideally should be in different commits. Updating the `cbor2` module is also a separate thing.

While I would like machine translations to be generated in pull requests (https://github.com/liberapay/liberapay.com/issues/75#issuecomment-3396512702), it should be done automatically with an agreed upon tool, not in unspecified impromptu ways that are likely to introduce inconsistencies and merge conflicts, so you should remove them from this PR.

New user secrets should go into the existing `user_secrets` database table or new separate tables, not in a weird `user_2fa` table. Moreover, any database schema changes are supposed to go into `sql/branch.sql`, not directly into the migration and schema files.

8by3:
@Changaco Ok, thanks for the pointers, I'll drop the translations and split 2fa into OTP / webauth PRs. Will raise tonight when I have the time.

PR #2802, opened 2026-08-26
Add Forgejo support (linked accounts + repositories)

pat-s:
Adds Forgejo (a Gitea-compatible, self-hostable git forge) as a supported platform for both "Linked accounts" (Elsewhere) and "Repositories" import.

## Approach

Forgejo is self-hostable with no single canonical instance, so it is modelled as a multi-domain OAuth2 platform, reusing the existing Mastodon/Pleroma machinery (the `username@domain` connect form, the `/on/forgejo:{domain}/associate` callback flow, and the per-instance `oauth_apps` credentials table).

Unlike Mastodon, Forgejo/Gitea has no dynamic OAuth-app registration API, so credentials cannot be created programmatically.
The new `Forgejo.register_app()` therefore does not call any API; it raises a friendly `LazyResponse(502)` for instances that have not been configured.
Credentials are seeded by the operator instead.

Repository listing is done unauthenticated against `/users/{user_name}/repos` (`use_session=False`), mirroring GitLab's public-only privacy stance.

Scope is individual accounts and their repositories (`has_teams = False`); organizations/teams are left for a follow-up.

## Operator setup (per Forgejo instance)

No schema migration is needed (the `oauth_apps` table already exists).
To enable an instance, a site administrator creates an OAuth2 application on that instance with the redirect URI `https://liberapay.com/on/forgejo:<instance-domain>/associate`, then inserts the resulting credentials:

```sql
INSERT INTO oauth_apps (platform, domain, key, secret)
VALUES ('forgejo', '<instance-domain>', '<client_id>', '<client_secret>');
```

There is no way around this manual step, as Forgejo has no endpoint for self-registering OAuth applications that could be used to store the retrieved credentials and onboard instances through the UI.

This setup is also documented in the `liberapay/elsewhere/forgejo.py` module docstring.

## Incidental fixes

Two small fixes are included because the Forgejo platform needs them:

- `header_links_paginator` ignored its `total_header` argument and always read the hardcoded `X-Total` header. Forgejo returns `X-Total-Count`.
- `AppConf` in `wireup.py` read `self.__annotations__`, which no longer falls through to the class under PEP 749 (Python 3.14), and called `isinstance()` on annotations that can be parameterized generics. Both are now handled.

Happy to split either of these into its own PR if you'd prefer.


martianhi:
@pat-s maybe i misunderstand your comments (i know v little about auth, tho i implemented a forgejo client), but isn't `POST /user/applications/oauth2` a way of generating oauth credentials programmatically?

e.g. https://codeberg.org/api/swagger#/user/userCreateOAuth2Application

in any case, great to see this being added.

pat-s:
Yes the endpoint exists but it needs authentication, which is what makes it unusable here.

`POST /user/applications/oauth2` sits under the `/user` API group, so it requires an access token for a user account on that instance (`admin:application` scope), and the app it creates is owned by that user.  
(Mastodon's `POST /api/v1/apps`, by contrast, is anonymous.)

That matters because of where the hook lives in Liberapay: `PlatformOAuth2.get_credentials()` calls `register_app(domain)` with nothing but a domain name (at the moment someone first types `me@<platform URL>`), before anyone has authenticated to that instance. So there's a chicken-and-egg problem: we'd need credentials on the instance to create the credentials.

The only workaround would be asking the connecting user for a PAT with `admin:application` scope, which is a lot of privilege to hand over just to link an account, and would leave the app owned by one arbitrary user who could revoke it for everyone else on that instance.  Therefore, seeding a row in `oauth_apps` per instance seemed like the saner trade.

martianhi:
Thanks for clarifying!

Patrick Schratz ***@***.***> writes:

> *
> pat-s left a comment (liberapay/liberapay.com#2802)
>
> Yes the endpoint exists but it needs authentication, which is 
> what makes it unusable here.
>
> POST /user/applications/oauth2 sits under the /user API group, 
> so it requires an access token for a user account on that 
> instance
> (admin:application scope), and the app it creates is owned by 
> that user.
> (Mastodon's POST /api/v1/apps, by contrast, is anonymous.)
>
> That matters because of where the hook lives in Liberapay: 
> PlatformOAuth2.get_credentials() calls register_app(domain) with
> nothing but a domain name (at the moment someone first types 
> me@<platform URL>), before anyone has authenticated to that 
> instance.
> So there's a chicken-and-egg problem: we'd need credentials on 
> the instance to create the credentials.
>
> The only workaround would be asking the connecting user for a 
> PAT with admin:application scope, which is a lot of privilege to 
> hand
> over just to link an account, and would leave the app owned by 
> one arbitrary user who could revoke it for everyone else on that
> instance. Therefore, seeding a row in oauth_apps per instance 
> seemed like the saner trade.
>
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> Triage notifications, keep track of coding agent tasks and 
> review pull requests on the go with GitHub Mobile for iOS and 
> Android.
> Download it today! 
> You are receiving this because you commented.


-- 
- Marty

pat-s:
@Changaco

Would be great to get a review here - from you or any other maintainer.

